Issue #63, opened 2015-05-08
Specifying separate dial host and http "Host:" header


pwaller:
I'm making a proxy which can handle web sockets. Everything is fine except I can't easily specify the `Host:` header correctly.

The problem is, some random client connects inbound with `example.com`, which proxies to `foo.internal`. `foo.internal` ends up seeing `Origin: example.com` and `Host: foo.internal`. I want the internal server to see `Host: example.com`.

This can be achieved in two ways I can immediately think of. I could implement my own Dial method:

https://github.com/gorilla/websocket/blob/ecff5aabe41f13b4cdf897e3c0c9bbccbe552a29/client.go#L182-L227

However, I would have to copy a great deal of the code in your library, and would not get the benefit of any fixes that might arrive in the future.

An alternative would be to extend the gorilla `Dial` method so that if a `Host` header is specified via the requestHeader parameter, it modifies the host of the URL `u` that is passed to `NewClient` (after the TCP dial has already happened), and gets deleted from the set of headers (so that it doesn't end up specified twice).

Would the latter fix be accepted as a PR, or can you think of an alternative method of achieving my goal? Thanks!

pwaller:
I've made a pull request. The change is quite small but took me a while to work out exactly what it should be, but I'm now reasonably sure this has the semantics I need. I've tested it in a production-like configuration and I'm happy with it.

garyburd:
An application can connect to a host different from the one specified in the URL using [Dialer.NetDial](http://godoc.org/github.com/gorilla/websocket#Dialer.NetDial):

      d := websocket.Dialer{
            NetDial: func(network, addr string) (net.Conn, error) {
                return new.Dial(network, proxyHostPort)
           }
     }
     ws, resp, err := d.Dial(u, h) // connects to proxyHostPort with Host header from u.

pwaller:
I think I discounted that approach because of the `ServerName` in the TLS config. I want the `ServerName` to be the name of the machine I'm connecting to, not the machine I'm proxying (and for that to be arranged for me by the Dial function.

I now see that I can arrange this if I need to by setting a TLS configuration. However this is less convenient than specifying the host header (which also makes the intent of my code clear - connect to this URL, use this host header).

I'll understand though if you don't want to accept my pull request. I'll leave you to close this issue and choose whether to close or accept my pull request.

Thanks for your time the last couple of days.

garyburd:
Fixed by 6fd0f867fef40c540fa05c59f86396de10a632a6

PR #64, opened 2015-05-08
Allow separate specification of http/dial Hosts

pwaller:
Enable the `Host:` header to be customized in a `Dialer.Dial` call.

This is needed to implement websocket proxies.

Please see discussion in #63.

### DIFF
--- a/client.go
+++ b/client.go
@@ -228,6 +228,22 @@ func (d *Dialer) Dial(urlStr string, requestHeader http.Header) (*Conn, *http.Re
 		requestHeader = h
 	}
 
+	if len(requestHeader["Host"]) > 0 {
+		// This can be used to supply a Host: header which is different from
+		// the dial address.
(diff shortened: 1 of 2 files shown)


garyburd:
Use this test for checking for the host header:

       if  len(requestHeader["Host"]) > 0 [
       
This handles the case where request header is map[string][]string{"Host": {}}

Do not modify the caller's request header. 

Add a test.

Update the CL description to describe the change.  The justification and link to discussion is nice to have, but not required.

pwaller:
Done. Test added, request header not modified, condition fixed. Thanks again for considering this.

I wasn't sure what you wanted on "Update the CL description to describe the change". I added the phrase to commit and PR body:

> Enable the `Host:` header to be customized in a `Dialer.Dial` call.

Let me know if that doesn't suffice.

PR #79, opened 2015-09-17
Ignore other previously sent headers in upgrade

zombiezen:


garyburd:
It's an error for the application to specify the headers added in this PR.  Upgrade should return an error instead of ignoring them.

garyburd:
Fixed by a4e0143e1f01fed44746a0175ccdb4926365aedd.

PR #89, opened 2015-10-18
Fix to remove possible race condition with initialisation of Dialer.

wolfeidau:
Moved the initialisation of the `Dialer` to the declaration of `DefaultDialer` to avoid possible race condition.

Credit goes to @davecheney for spotting the issue.

### DIFF
--- a/client.go
+++ b/client.go
@@ -154,7 +154,7 @@ func hostPortNoPort(u *url.URL) (hostPort, hostNoPort string) {
 }
 
 // DefaultDialer is a dialer with all fields set to the default zero values.
-var DefaultDialer *Dialer
+var DefaultDialer = &Dialer{}
 
 // Dial creates a new client connection. Use requestHeader to specify the
 // origin (Origin), subprotocols (Sec-WebSocket-Protocol) and cookies (Cookie).


davecheney:
Hmm, on reflection I think I am wrong, there is no race, `d` being updated here is the local copy, not the package level var.

With that said, this lazy initalisation is a bit odd, why not declare `DefaultDailer` as a value, not a pointer?

garyburd:
There is no race.

I declared the default value as a pointer type because [a similar variable in the standard library](https://godoc.org/net/http#DefaultClient) is a pointer type. Changing the type of the variable will break API compatibility.

I have no explanation for why the variable is set to nil. Setting it to &Dialer{} seems better.  I will accept a PR for this change.

davecheney:
Yup, that'll work.

wolfeidau:
Did you want me to close this and open a new pull request with a more accurate commit message or just update the commit message on this one?

davecheney:
Dunno, does you religion permit you to use --amend and --push ?

garyburd:
Do not remove the nil check. I prefer not to break apps that depend on it.

wolfeidau:
OK commit that just changes the one line is now in with a different commit message.

Cheers.

PR #90, opened 2015-10-20
Proxy support for websocket clients.

wolfeidau:
Firstly the intent of this pull request is to enable websocket clients to traverse proxy servers using the standard environment variables in line with the golang http package.

The reason I am submitting this patch is one of the upstream consumers of this library currently doesn't work when run within an isolated network environment (see https://github.com/aws/amazon-ecs-agent/issues/23).

I have tried to incorporate a simple and easily overridable API along with test coverage for my additions.

### DIFF
--- a/client.go
+++ b/client.go
@@ -5,6 +5,7 @@
 package websocket
 
 import (
+	"bufio"
 	"bytes"
 	"crypto/tls"
 	"errors"
@@ -48,6 +49,12 @@ type Dialer struct {
 	// NetDial specifies the dial function for creating TCP connections. If
 	// NetDial is nil, net.Dial is used.
 	NetDial func(network, addr string) (net.Conn, error)
+
+	// Proxy specifies a function to return a proxy for a given
+	// Request. If the function returns a non-nil error, the
+	// request is aborted with the provided error.
+	// If Proxy is nil or returns a nil *URL, no proxy is used.
+	Proxy func(*http.Request) (*url.URL, error)
 
 	// TLSClientConfig specifies the TLS configuration to use with tls.Client.
 	// If nil, the default configuration is used.
@@ -110,9 +117,12 @@ func hostPortNoPort(u *url.URL) (hostPort, hostNoPort string) {
 	if i := strings.LastIndex(u.Host, ":"); i > strings.LastIndex(u.Host, "]") {
 		hostNoPort = hostNoPort[:i]
 	} else {
-		if u.Scheme == "wss" {
+		switch u.Scheme {
+		case "wss":
 			hostPort += ":443"
-		} else {
+		case "https":
+			hostPort += ":443"
+		default:
 			hostPort += ":80"
 		}
 	}
@@ -120,7 +130,9 @@ func hostPortNoPort(u *url.URL) (hostPort, hostNoPort string) {
 }
 
 // DefaultDialer is a dialer with all fields set to the default zero values.
-var DefaultDialer = &Dialer{}
+var DefaultDialer = &Dialer{
+	Proxy: http.ProxyFromEnvironment,
+}
 
 // Dial creates a new client connection. Use requestHeader to specify the
 // origin (Origin), subprotocols (Sec-WebSocket-Protocol) and cookies (Cookie).
@@ -134,7 +146,9 @@ var DefaultDialer = &Dialer{}
 func (d *Dialer) Dial(urlStr string, requestHeader http.Header) (*Conn, *http.Response, error) {
 
 	if d == nil {
-		d = &Dialer{}
+		d = &Dialer{
+			Proxy: http.ProxyFromEnvironment,
+		}
 	}
 
 	challengeKey, err := generateChallengeKey()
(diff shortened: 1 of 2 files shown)


garyburd:
- Structure the code so the [net/http ProxyFromEnvironment](https://godoc.org/net/http#ProxyFromEnvironment) function can be used as is. 

- Specify proxy function using Dialer field instead of package level variable.  As with net/http, the default is no proxy.

        // Proxy specifies a function to return a proxy for a given
        // Request. If the function returns a non-nil error, the
        // request is aborted with the provided error.
        // If Proxy is nil or returns a nil *URL, no proxy is used.
        Proxy func(*http.Request) (*url.URL, error)

- I will submit a PR in a few hours to move the code from NewClientConn to Dialer.Dial and rewrite NewClientConn using Dialer.Dial.  With this change, the proxy code can use the  connection's [buffered reader](https://github.com/gorilla/websocket/blob/423912737d2c26382ab4b2647d796811d4efa00b/conn.go#L161) to read the response from the proxy server.

garyburd:
See refactored code at https://github.com/garyburd/websocket/commit/ae2bbac1c1729aea27efd3cb2c168c08becf572a.  I want to add some tests before pushing it to master.  

The the net/http DefaultTransport uses ProxyFromEnvironment.  The websocket DefaultClient should as well.

wolfeidau:
Just a note you don't want to use a HTTP proxy for websockets as it is just a primitive request relay, websockets will need to always use the HTTPS proxy with the CONNECT method.

This is one of the reasons I forked and adjusted the `ProxyFromEnvironment`.

That said more alignment between would be idea!

garyburd:
What's the reason for not using the CONNECT method on an HTTP_PROXY?

wolfeidau:
Good point they will both typically be the same these days, although a lot of the older proxies still have ACLs for either HTTP or HTTPS both are always permitted.

garyburd:
The main issue is deciding on the protocol to use when calling the `Proxy func(*http.Request) (*url.URL, error)` function. Should the package pass the handshake request as is to the function or should the package always set the protocol to "https"?   

Hopefully we don't need the forked code. I don't want to figure how to bring code in the project with a different license.

wolfeidau:
I absolutely agree on the licence side, I can understand why your loath to add the code as it is, that is why I made the intent of the change very clear.

I would recommend forcing use of the HTTPS proxy just to be more "correct", if that requires tweaking the URL to convince that method to give you the URL of HTTPS Proxy I think that is OK. 

Just my two cents.

garyburd:
The Dial refactoring is pushed to master.  Please do the following:

- Add Proxy field to Dialer.

        // Proxy specifies a function to return a proxy for a given
        // Request. If the function returns a non-nil error, the
        // request is aborted with the provided error.
        // If Proxy is nil or returns a nil *URL, no proxy is used.
        Proxy func(*http.Request) (*url.URL, error)

- Set DefaultDialer.Proxy to http.ProxyFromEnvironment.

- Write the test following the structure of TestHostHeader.  Instead of defining another test server type, [hook the handler in the existing test server](https://github.com/gorilla/websocket/blob/5ed2f4547d7b9c036541b27a67ecf54e41756997/client_server_test.go#L299-L306).

client.go
```diff
@@ -220,6 +274,32 @@ func (d *Dialer) Dial(urlStr string, requestHeader http.Header) (*Conn, *http.Re
...
+			Host:   hostPort,
+			Header: make(http.Header),
+		}
+
+		connectReq.Write(netConn)
+
+		// Read response.
+		// Okay to use and discard buffered reader here, because
+		// TLS server will not speak until spoken to.
+		br := bufio.NewReader(netConn)
```
garyburd: Use c.br. No need to create another one.

client.go
```diff
@@ -220,6 +274,32 @@ func (d *Dialer) Dial(urlStr string, requestHeader http.Header) (*Conn, *http.Re
...
+		resp, err := http.ReadResponse(br, connectReq)
+		if err != nil {
+			netConn.Close()
```
garyburd: Deferred function will close netConn. Don't close it here.

client.go
```diff
@@ -220,6 +274,32 @@ func (d *Dialer) Dial(urlStr string, requestHeader http.Header) (*Conn, *http.Re
...
+			return nil, nil, err
+		}
+		if resp.StatusCode != 200 {
+			f := strings.SplitN(resp.Status, " ", 2)
+			netConn.Close()
```
garyburd: Deferred function will close netConn. Don't close it here.

client_server_test.go
```diff
@@ -123,6 +123,47 @@ func sendRecv(t *testing.T, ws *Conn) {
...
 }
 
+func TestProxyDial(t *testing.T) {
+
+	s := newServer(t)
+	defer s.Close()
+
+	surl, _ := url.Parse(s.URL)
+
+	cstDialer.Proxy = func(r *http.Request) (*url.URL, error) {
```
garyburd: cstDialer.Proxy = http.ProxyURL(surl)

client.go
wolfeidau: I had a look at doing this but the `conn` isn't created till after the TLS handshake, which matches the core API, hence the need for the `br` at this stage.

wolfeidau:
Latest round of fixes is up and tested with the caveat / question / comment above.

Cheers

client.go
garyburd: Ah, got it.

garyburd:
I was informed that client.go contains copyrighted code. I need to figure out what to do with the licensing.

wolfeidau:
I can rewrite the helper methods I have used from the golang API if required, this use case is bit more specific anyway.

garyburd:
Replace use of canonicalAddr and supporting functions with existing function hostPortNoPort.  Update hostPortNoPort to handle "http" and "https".

wolfeidau:
Sterling idea, done, tested and pushed.

PR #92, opened 2015-10-24
chat example revised to use single goroutine per connection

mcqueenorama:
The chat example seems to allow multiple goroutines reading from connections in an unsafe way.  This example does one goroutine per connection.

### DIFF
--- a/examples/chat/conn.go
+++ b/examples/chat/conn.go
@@ -5,13 +5,15 @@
 package main
 
 import (
-	"github.com/gorilla/websocket"
 	"log"
 	"net/http"
 	"time"
+
+	"github.com/gorilla/websocket"
 )
 
 const (
+
 	// Time allowed to write a message to the peer.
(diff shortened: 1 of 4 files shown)


garyburd:
The example does not allow multiple goroutines to read from a single connection.

The `readPump` function is the only function that reads from the connection.  The readPump function is called once per connection [from the handler function](https://github.com/gorilla/websocket/blob/527637c0f38a8035d7856bb758e56f7ee2c704a9/examples/chat/conn.go#L105).

garyburd:
Please file an issue describing how the example reads concurrently from a single connection. If there is a bug, I'll implement the fix.

This PR is not acceptable because it leaks tickers, writes to connections concurrently and adds a name to the broadcast channel for no apparent reason.

mcqueenorama:
Thanks for taking the time to check out the code and explain.  Its a cool project.

PR #93, opened 2015-10-30
Saticfy go lint.

serjvanilla:
if block ends with a return statement, so drop this else and outdent its block.


garyburd:
Thank you for pointing this out.  The code in this PR is not gofmt'ed. I pushed 044091b51d3b40b9e7fbc1fb34975a7662916e4e to fix the issue.

PR #94, opened 2015-11-01
Allow cross-origin requests for example

MrDunne:
For the sake of testing the websocket server, accepting cross-origin requests is useful - to allow usage of things like http://www.websocket.org/echo.html

### DIFF
--- a/examples/echo/server.go
+++ b/examples/echo/server.go
@@ -20,7 +20,8 @@ var upgrader = websocket.Upgrader{
 	CheckOrigin: func(r *http.Request) bool {
         return true
     },
-} // use default options, except origin. Allows cross-origin requests for testing.
+} /* use default options, except origin. Allows cross-origin requests for testing. 
+Presents a security risk if run in production */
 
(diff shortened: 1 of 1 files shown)


garyburd:
The change as is does not support the goal of the example and can be harmful to copypaste programmers.

If you want to test something like http://www.websocket.org/echo.html, then let's add that to the example server.

MrDunne:
The point was more that it allows people to test it easily, and is that not what I did? It's already an echo server but the only issue was the cross origin. However, would it meet both our goals with this latest commit?

garyburd:
The goal of the example is to show how write a basic client and server.  It is not a goal to provide an echo server. I am open to changes that improve the code as an example.

MrDunne:
Sure, I mean I can't think of many situations where an echo server provides much utility. The motivation behind it was simply to make it easy for someone just looking to test out the framework quickly

garyburd:
The example now includes functionality similar to http://www.websocket.org/echo.html.

PR #99, opened 2015-12-16
Handle query strings properly when parsing url

meinside:
Query strings were not handled properly while parsing url.


garyburd:
Do you have a case where the client does not correctly write the query string to the network?

meinside:
Never mind. It was due to wrongly generated url string from manually built url.URL.

PR #102, opened 2015-12-17
fix case sensitive header check

### LINKED ISSUE #101: Wrong case for Sec-WebSocket-Version
According to [RFC 6455](https://tools.ietf.org/html/rfc6455#section-11.3.5) the correct header for specifying the web socket version is `Sec-WebSocket-Version`, instead of `Sec-Websocket-Version` as it is [checked by the upgrader](https://github.com/gorilla/websocket/blob/567453a710c89724d4e971da34f613b0b89961cc/server.go#L99).

This means connection upgrading fails for requests that specify the correct header.

pjvds:
The upgrader is checking the `Sec-Websocket-Version` in a case sensitive way directly against the header map. This means, if someone sends the correct header as specified by [RFC 6455](https://tools.ietf.org/html/rfc6455#section-11.3.5) the code fails. The fix checks the header with the canonical formats.

This should fix #101

### DIFF
--- a/server.go
+++ b/server.go
@@ -96,7 +96,7 @@ func (u *Upgrader) Upgrade(w http.ResponseWriter, r *http.Request, responseHeade
 	if r.Method != "GET" {
 		return u.returnError(w, r, http.StatusMethodNotAllowed, "websocket: method not GET")
 	}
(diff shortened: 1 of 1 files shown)


garyburd:
Please reopen if you can provide information on how to reproduce issue #101.

PR #110, opened 2016-02-16
Close Frame Response

tabone:
Taking a look at the documentation of the [Close Frame](https://tools.ietf.org/html/rfc6455#section-5.5.1) in the WebSocket RFC it mentions the following:

> When sending a Close frame in response, the endpoint typically echos the status code it received.

This PR does just that, when a Close Frame is received, it echoes it to the other endpoint.

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -726,13 +726,15 @@ func (c *Conn) advanceFrame() (int, error) {
 			return noFrame, err
 		}
 	case CloseMessage:
-		c.WriteControl(CloseMessage, []byte{}, time.Now().Add(writeWait))
+		echoMessage := []byte{}
 		closeCode := CloseNoStatusReceived
 		closeText := ""
 		if len(payload) >= 2 {
+			echoMessage = payload[:2]
 			closeCode = int(binary.BigEndian.Uint16(payload))
 			closeText = string(payload[2:])
 		}
+		c.WriteControl(CloseMessage,  echoMessage, time.Now().Add(writeWait))
 		return noFrame, &CloseError{Code: closeCode, Text: closeText}
 	}
 


garyburd:
The package does not echo the status code because it makes it easier to debug close issues at the peer. As currently written, the peer does not need to distinguish between a close frame indicating an error on the part of the peer and an echo of an error.

Does the proposed change solve a problem or provide a benefit?

garyburd:
I looked at other servers and found that they echo the status code. I agree to follow along.

The code to echo the status code should look something like this:

                echoMessage := []byte{}
		if len(payload) >= 2 {
			closeCode = int(binary.BigEndian.Uint16(payload))
			closeText = string(payload[2:])
                       echoMessage = payload[:2]
		}
		c.WriteControl(CloseMessage,  echoMessage, time.Now().Add(writeWait))

I am still curious to know if the proposed change solved a problem or provided some benefit.

tabone:
To tell you the truth I wasn't aware that this package resends a `Close Frame` with code 1005 once it receives one and therefore I tried to implement it in my application. I found it strange that the `Close Frame` being sent as a response wasn't the one I was sending and thus began looking at your code Currently I couldn't tell you whether I'll benefit from it, but since the WebSocket RFC mention it, I thought it would be nice to contribute.

garyburd:
Please squash the commits.

garyburd:
> I found it strange that the Close Frame being sent as a response wasn't the one I was sending 

You should not depend on the response being the same as the one you are sending:
- The RFC does not require or recommend that an endpoint echo the status code.  
- Even if the peer is written to echo the status code, it's possible the peer initiates the close handshake on its own.

tabone:
Indeed and thanks for the points you listed. I found it strange because in my prototype I was trying to echo the same code as response and instead I was always receiving 1005.

PR #133, opened 2016-05-18
Example code clarity improvement

jadekler:
At the moment it's a little hard to figure out what's going on when a bunch of variables are `h`, `c`, `m`, and so on. This is pretty normal shorthand Go users, but in example code it's quicker to understand if shorthand is avoided.

edit: Happy to refactor the names to something else if you prefer. `mainHub` especially I don't love.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,3 +20,6 @@ _cgo_export.*
 _testmain.go
 
 *.exe
+
+.idea/
+*.iml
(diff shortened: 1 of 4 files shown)


garyburd:
Keep the single character receiver names.  I don't want to stray too far from the Go style guidelines.  

The name "hub" is a better name than "mainHub".

PR #136, opened 2016-05-28
Add Ping/Pong handler Getter functions

reeze:
When set custom handler we must reply to ping message with pong manually. 

This allow us to get the original handler and chain the handlers.


garyburd:
Name the methods PingHandler and PongHandler to match Go naming conventions. Squash the commits.

reeze:
DONE.  Thank you.

PR #150, opened 2016-07-11
Update hub.go

RHavar:


garyburd:
Does this fix an issue?

RHavar:
The example only has one instance of `Hub` so it doesn't really matter, but if you created a second hub the `run` method would not work because it's (accidentally?) hardcoded to the `hub` instance.

garyburd:
Ouch, it was accidentally hardcoded to the single hub instance. Thanks!

PR #164, opened 2016-10-13
Update README to reflect change of Go sub-repo websocket feature

artyom:
Reflect recent update to golang.org/x/net/websocket that introduced limit of received messages.


garyburd:
The table row should be deleted.

artyom:
Please see updated commit.

PR #166, opened 2016-10-17
add cookie jar to dialer

MaximeHeckel:
This allows users to add a cookie jar the same way it can be done for the `http.Client` of the `net/http` package

Example:

```
import (
	"net"
	"net/http"
	"net/http/cookiejar"

)

func main () {
    var jar http.CookieJar
    jar, _ = cookiejar.New(nil)

    Url := "http://example.com"
    Dialer := websocket.Dialer{Jar: jar}
    header := http.Header{}
    ws, res, err := Dialer.Dial(Url, header)

    ...
}
```

### DIFF
--- a/client.go
+++ b/client.go
@@ -78,6 +78,11 @@ type Dialer struct {
 	// guarantee that compression will be supported. Currently only "no context
 	// takeover" modes are supported.
 	EnableCompression bool
+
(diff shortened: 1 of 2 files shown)


MaximeHeckel:
I added the cookies of the jar in the response, and a test

PR #173, opened 2016-10-25
Add mutex to prevent concurrent connection write

powelles:


garyburd:
Please describe the data race that this fixes.

powelles:
I just looked at the code in detail and it appears you may have fixed the issue I was having. Using an earlier version I could get `panic: concurrent write to websocket connection` from the writePump.

garyburd:
Do you know the time git commit or the approximate time you pulled the code that produced the panic?

powelles:
I was just able to replicate it with 0b847f2facc24ec406130a05bb1bb72d41993b05 using the older style relying on `conn.WriteMessage`

garyburd:
Just so we are on the same page here, are you saying that the chat example in  0b847f2facc24ec406130a05bb1bb72d41993b05 panics?  I cannot reproduce it.

powelles:
The chat example as is doesn't panic as far as I'm aware. I adapted the example and call the `Client` method `write(int, []byte) error` from different goroutines and that is where I'm getting a panic. I though it might do someone else a favor to make it where that method can't cause a panic. At least not in the same way that I was getting one.

garyburd:
The [documentation](https://godoc.org/github.com/gorilla/websocket#hdr-Concurrency) describes the concurrency supported by a connection.  

The chat example ensures there's no more than one concurrent writer by using a single goroutine for all writes to the connection. There's no need the mutex.

powelles:
After your first response I realized I jumped the gun a little with the pull request. The previous version of the chat example (which my code was based on) had the issue my pull request was designed to fix. I'm sorry if I didn't make that clear. I do think it might be helpful to make it more clear in some way that the example client api isn't safe to use.

powelles:
Saw your commit. Thanks. Hopefully that will help someone from making my mistake.

PR #176, opened 2016-10-27
Don't close conn/FD on write errors

KiranRG:
We have built a "streaming" layer atop gorilla-websocket, that essentially buffers messages before writing out to the wire .. and on the other end, we would "deserialize" the payload into individual messages before returning to the application.

One of the issues we encountered when using gorilla's websocket implementation was in the following scenario -- we have a streaming client and a streaming server, each with read and write "pumps"; the client sends "requests" to the the server, which the server receives/processes and sends back "responses" to. After the server has sent the last of its "responses" (including a final "done" response) it closes the connection.

The client has a write-pump (that's sending requests) and a read-pump (that's reading/receiving responses). Because the server has gone away, the client's write-pump encounters an error when doing a write, and the gorilla layer (in the code-path that I have fixed below) automatically/internally closes the connection/FD -- this effectively wipes out any buffered responses (including the "done" response) from the server, which would have otherwise been available to read.

In our scenario, the client re-connects to the server and retries until it gets the "done" response -- which it fails to see due to the above bug, causing it to be stalled doing retries for a long/indefinite time.

What this fix does is to avoid closing the connection on a write-error -- we leave it to the application/user (of gorilla websocket) to call websocket.Close() when done. In fact, for errors from conn.Write, the current code does simply convey it to the user, but unfortunately the check for a partial-write seems to take precedence (and this check would succeed even for a write-failure), causing it to close the connection any time there is a write-failure.

In our case, for instance, this gives an opportunity for our client's read-pump to be able to "drain" out all available messages ("responses", including the "done" response), before encountering the error (due to the connection closure) and closing the connection.

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -177,6 +177,7 @@ var (
 	errBadWriteOpCode      = errors.New("websocket: bad write message type")
 	errWriteClosed         = errors.New("websocket: write closed")
 	errInvalidControlFrame = errors.New("websocket: invalid control frame")
+	errPartialWrite        = errors.New("websocket: partial write")
 )
 
 func hideTempErr(err error) error {
@@ -322,12 +323,11 @@ func (c *Conn) write(frameType int, deadline time.Time, bufs ...[]byte) error {
 	for _, buf := range bufs {
 		if len(buf) > 0 {
 			n, err := c.conn.Write(buf)
-			if n != len(buf) {
-				// Close on partial write.
-				c.conn.Close()
-			}
 			if err != nil {
 				return err
+			}
+			if n != len(buf) {
+				return errPartialWrite
 			}
 		}
 	}
@@ -387,8 +387,8 @@ func (c *Conn) WriteControl(messageType int, data []byte, deadline time.Time) er
 
 	c.conn.SetWriteDeadline(deadline)
 	n, err := c.conn.Write(buf)
-	if n != 0 && n != len(buf) {
-		c.conn.Close()
+	if err == nil && n != 0 && n != len(buf) {
+		err = errPartialWrite
 	}
 	return hideTempErr(err)
 }


garyburd:
The server application should send a close frame instead of closing the connection. Close the connection on errors only.

KiranRG:
I missed mentioning that, but yes, our server does send a "close frame" just before closing the connection. But again, that is not seen by the client's read-path .. because when the the client write fails, gorilla just closes the connection. So the error we see on the client's read path is always "abnormal closure" .. not the graceful "going away".

After this fix, we see the "going away" error (from the close frame).

garyburd:
The server application should not close the connection.

KiranRG:
That said, the semantics we define is that when the server is done, it could go away any time -- and stop accepting "requests". But if the client needs to go away, it should, in the graceful scenario, wait until it reads out all the "responses" for the requests that were previously sent are received; or take the cost of reconnecting, sending requests and getting the responses for them again.

I know these are semantic differences and it is arguable when one is better than the other. But that aside, if you look at the code where this fix is, the gorilla code is checking on the _length_ of the returned message even before it checks to see if the write failed or not; and this check would succeed even when the write has failed, making the write-failed-check moot.

Also, as the code is currently, when the write has actually failed, we do not close the connection -- which is inconsistent with the partial write case (though I think is the right thing to do). Any errors returned by this function are stored in the 'writeErr', causing any future writes to fail anyway. So I think it is okay to leave the connection open .. so at least the reads can be drained.

If you'd like I could write up a sample/test to explain the scenario better?

garyburd:
I agree that there's an issue with the code in WriteControl. The test for number of bytes written should be replaced with err != nil.  It may not be necessary to close the connection at all. I need more time to review the code to understand why the connection is closed.

The value errPartialWrite in the PR will never be used because an io.Writer always returns a non-nil error on partial write.

A server application should implement graceful close as follows:

1. Send a close frame
2. Continue reading until read returns an error or a grace period expires.
3. Close the connection

If the server closes gracefully, then the client will not encounter a write error (including partial writes).

KiranRG:
Thanks Gary.

Yes, I wasn't sure if it is really possible for a conn.Write to succeed (ie, return nil error), when only a partial write (ie 'n != len(buf)') has occurred. We probably could do away with the partial write check altogether -- but propagate any errors from the write; but again, _without_ closing the connection.

With regards to the "graceful close" from the server, I understand there's more than one way to define a "graceful close" exchange -- but I think the websocket implementation should, as far as possible, **not** have to impose a policy on the application, would you agree?

Btw, why we do not prefer the server to have to wait to hear back from the client is, like I mentioned, to avoid holding back resources (network connections, internal buffers, etc) when it knows it has nothing more for the client. So when done, the server just flushes out all the messages it has, sends an "end" message (defined by our protocol) and a websocket close-frame before closing out the connection. The client can take its own time to read all of the messages before seeing the end of the stream. If the websocket layer does not close the connection/FD on a write error, we are able to achieve this behaviour.

garyburd:
The handshake using close frames is specified by the WebSocket RFC.

The package closes the network connection on partial writes to force subsequent writes to return an error.  If the connection continues after a partial write, then the stream will be out of sync between the endpoint and its peer.

An alternative approach is to latch the error in WebSocket connection.  The latch can be combined with the existing CloseSent field.

KiranRG:
Gary, the error returned by the 'write(..)' method (where I have the proposed fix) is called by 'flushFrame()' and is stored in 'Conn::writeErr'. Every write method (Write(), NextWriter(), etc) checks this field to ensure it is non-nil before allowing further writes. So this mechanism already ensures all future writes are fail.

Again, not closing the FD (on write failure) helps the application be able to drain out all available messages that can be read -- and being able to do so is crucial in our scenario.

garyburd:
An error in WriteControl is not stored in writeErr.

garyburd:
I'll work on this.

KiranRG:
Thanks Gary.

PR #195, opened 2016-12-17
pool flate readers

y3llowcake:


garyburd:
Add code to return the flate reader to the pool when the reader returns io.EOF.  This will help with scenarios where the application does not call  NextReader or there's a delay calling NextReader.

y3llowcake:
Added this check into flateReadWrapper.Read()

y3llowcake:
And it actually compiles now :grimacing:

PR #202, opened 2017-01-05
Add the SetDeadline method on the connection

ellimist:
I was using recently the deadlines and noticed that the SetDeadline method of the connection is missing.

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -292,6 +292,8 @@ func newConn(conn net.Conn, isServer bool, readBufferSize, writeBufferSize int) 
 	return c
 }
 
+// SetDeadline sets the read and write deadline on the underlying network connection
+// It is equivalent to calling both SetReadDeadline and SetWriteDeadline.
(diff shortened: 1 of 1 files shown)


garyburd:
Why does your application need this method?

garyburd:
This method is not the same as calling SetReadDeadline and SetWriteDeadline.

If your goal is to bypass SetWriteDeadline, then call c.UnderlyingConn().SetWriteDeadline().

If your goal is to set the read and write deadlines to the same value, then call both SetReadDeadline and SetWriteDeadline.  I do not want to add a helper method for this because the it's not something that most applications need. Also, the method is difficult to use correctly given the concurrency constraints on connections.

A websocket connection is not a net.Conn. There's no need add a method just because net.Conn has the method.

ellimist:
I understand now. Looks like I jumped to conclusions a bit too fast :) Thanks for your reply

Issue #215, opened 2017-02-18
New TLS (Go 1.8) fields removed when cloning TLS config


willeponken:
Hello,

In the latest Go release (1.8) several new callbacks, `GetClientCertificate`, `GetConfigForClient` and `VerifyPeerCertificate`, has been added to tls.Config (https://golang.org/pkg/crypto/tls/#Config).

These callbacks is removed by the cloneTLSConfig function: https://github.com/gorilla/websocket/blob/master/client.go#L397

Also, couldn't Config.Clone (https://golang.org/pkg/crypto/tls/#Config.Clone) be used instead and then just set `nil` for the `SessionTicketsDisabled` and `SessionTicketKey` fields?

__Edit__: I see now that Config.Clone too was added in 1.8, would it be a good idea to use a 1.8 build constraint? I don't really see another way to fix this.

willeponken:
Fixed in #216

PR #216, opened 2017-02-18
Clone new tls.Config fields for Go 1.8

willeponken:
In Go 1.8 several new fields (GetClientCertificate, GetConfigForClient and VerifyPeerCertificate) for tls.Config was added. These should also be cloned. #215


garyburd:
Use the [Clone](https://godoc.org/crypto/tls#Config.Clone) method added in Go 1.8.

willeponken:
There you go!

PR #222, opened 2017-02-27
Return reason with HTTP error response

kulti:


garyburd:
What is the problem you are trying to solve with this change?  I want to understand why change is needed before discussing the issues with with this change.

kulti:
When I try to upgrade HTTP connection to WebSocket, I want to receive more informative message if smth went wrong. At now I can see only "Bad Request" on client side, e.g.

garyburd:
Which client are you using and how does it display the response?

kulti:
Self-written client for self-written server. Client writes logs. Or I can use curl to test ws-server. It doesn't metter. When I see response with "Bad Request" or other errors - I want to see reason to understand and fix problem.

garyburd:
I will not accept the change as is because it leaks internal errors to the client.

Can you debug your client by logging the error returned from upgrade on the server?  The error string contains the reason text.

garyburd:
I am saying that the original PR can leak internal errors, not that all errors are internal errors.

If you can debug using the server logs, then why do you need this change?

kulti:
I can, but client-side developer cannot.

I found the following errors for Bad Request:
- 'upgrade' token not found in 'Connection' header"
- 'websocket' token not found in 'Upgrade' header
- unsupported version: 13 not found in 'Sec-Websocket-Version' header
- not a websocket handshake: `Sec-Websocket-Key' header is missing or blank

and I think all of them are communication errors. Consequently, client need to know them. Accordingly to RFC "the server SHOULD include an entity containing an explanation of the error situation" https://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html#sec10.4

kulti:
Ok. I understood your point. And I agree that this messages for developers, but I think, client side developer also need in Bad Request explanations. At now I can solve my issue using Error hook. Thank you.

PR #241, opened 2017-05-16
Set keepalive for TCP sockets

cannium:
Set keepalive flags for TCP sockets to help detect if connection is dead in bad network conditions.
At least this patch will do no harm.

### DIFF
--- a/server.go
+++ b/server.go
@@ -162,6 +162,12 @@ func (u *Upgrader) Upgrade(w http.ResponseWriter, r *http.Request, responseHeade
 	}
 	var brw *bufio.ReadWriter
 	netConn, brw, err = h.Hijack()
+	tcpConn, ok := netConn.(*net.TCPConn)
+	if ok {
+		// Set TCP keepalive for connections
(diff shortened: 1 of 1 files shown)


dvdplm:
For the ignorant among us, what good will it do? I.e. what are the conditions under which having it set to true helps detecting disconnects? Do clients need to turn it on also?

garyburd:
The net/http ListenAndServe and ListenAndServeTLS functions [enable TCP keep alive](https://github.com/golang/go/blob/8f36668178d07e6f1b979252c1c68ebe788d14be/src/net/http/server.go#L3064-L3065).

I question the importance of this PR. It does not handle end to end detection of dead clients or TLS connections.  Is there a way to find the underlying network connection when TLS is used?

If we go forward with this, it should be a Dialer and Upgrader option and not enabled by default.

cannium:
@garyburd Thanks for pointing it out.
@dvdplm I am the ignorant. Sorry for the disturb.
I'll close the PR.

PR #250, opened 2017-06-12
[Examples\Chat] Allow to free HTTP conn things

ovr:
Hey!

Looks like without it, We are getting leak

Thank


garyburd:
Write it as

	go client.writePump()		 	
 	go client.readPump()

instead of adding a new method. Gofmt the code.

garyburd:
Implemented in ea4d1f681babbce9545c9c5f3d5194a789c89f5b

PR #251, opened 2017-06-13
fix isWriting not needed, because conn.Write aleady do this.

hetao29:


garyburd:
The code deleted by this PR is not necessary, but it does help application developers find concurrency bugs with negligible overhead.  We will keep the code.

`conn.Write` does not detect concurrent writes by the application.

hetao29:
Yes, thanks very much!

PR #255, opened 2017-06-20
add IsWriting func for ws funs

hetao29:
1. add IsWriting() for ws test the status.
2. add read & write mutex


garyburd:
It looks like the goal of this PR is to remove the restrictions on concurrent readers and writers.  I want to see a high level design for removing these restrictions before commenting on any code.  This design should handle all of the methods with restrictions on concurrent calls (NextWriter, EnableWriteCompression, SetWriteDeadline, ...) as well as example code showing how removing the restrictions makes it easier to write an application.

PR #257, opened 2017-06-22
removing the hardcoded 127.0.0.1

venkat:
makes it easy to connect when running the code in a remote VM, for example.


elithrar:
We should not listen on all interfaces by default as it is a security risk. 

You can change the address:port by passing the `-addr` flag when running the example.

PR #262, opened 2017-07-07
distinguish Upgrader.Upgrade from Upgrade

fuzzybear3965:
The documentation regarding which Upgrade method/function is deprecated was a bit confusing. I updated the language to make it a bit clearer.

### DIFF
--- a/doc.go
+++ b/doc.go
@@ -6,9 +6,8 @@
 //
 // Overview
 //
-// The Conn type represents a WebSocket connection. Within the context of an
-// HTTP request handler, a server application calls the Upgrade method of an
-// Upgrader instance obtaining a pointer to a Conn:
+// The Conn type represents a WebSocket connection. A server application calls
+// the Upgrader.Upgrade method from an HTTP request handler to get a *Conn:
 //
 //  var upgrader = websocket.Upgrader{
 //      ReadBufferSize:  1024,
@@ -152,10 +151,9 @@
 // origin policy checking using the CheckOrigin field associated with the
 // Upgrader instance.
 //
-// By contrast, the deprecated package-level Upgrade function
-// does not perform origin checking. In this case is the application's
-// responsibility to manually check the Origin header before calling the
-// package-level Upgrade function.
+// The deprecated package-level Upgrade function does not perform origin
+// checking. The application is responsible for checking the Origin header
+// before calling the Upgrade function.
 //
 // Compression EXPERIMENTAL
 //
--- a/doc.go
+++ b/doc.go
@@ -6,9 +6,9 @@
 //
 // Overview
 //
-// The Conn type represents a WebSocket connection. A server application uses
-// the Upgrade function from an Upgrader object with a HTTP request handler
-// to get a pointer to a Conn:
+// The Conn type represents a WebSocket connection. Within the context of an
+// HTTP request handler, a server application calls the Upgrade method of an
+// Upgrader instance obtaining a pointer to a Conn:
 //
 //  var upgrader = websocket.Upgrader{
 //      ReadBufferSize:  1024,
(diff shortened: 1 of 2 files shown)


server.go
```diff
@@ -228,12 +228,14 @@ func (u *Upgrader) Upgrade(w http.ResponseWriter, r *http.Request, responseHeade
 	return c, nil
 }
 
-// Upgrade upgrades the HTTP server connection to the WebSocket protocol.
+// DEPRECATED - use websocket.Upgrader instead.
```
fuzzybear3965: It would be good to change this line to  `Deprecated: use websocket.Upgrader instead.` in order to conform to the official godoc recommendations.

doc.go
```diff
@@ -147,9 +147,15 @@
 //      CheckOrigin: func(r *http.Request) bool { return true },
 //  }
 //
-// The deprecated Upgrade function does not enforce an origin policy. It's the
-// application's responsibility to check the Origin header before calling
-// Upgrade.
+// We recommend the upgrader.Upgrade method to perform an upgrade
+// from an HTTP connection to a websocket connection. This method performs
```
garyburd: The proceeding paragraphs in this section describe how CheckOrigin is used.  There's no need to repeat that the upgrader uses CheckOrigin.

doc.go
```diff
@@ -147,9 +147,15 @@
 //      CheckOrigin: func(r *http.Request) bool { return true },
 //  }
 //
-// The deprecated Upgrade function does not enforce an origin policy. It's the
-// application's responsibility to check the Origin header before calling
-// Upgrade.
+// We recommend the upgrader.Upgrade method to perform an upgrade
```
garyburd: There's no need to recommend the Upgrader.Upgrade. It's the only way non-deprecated API for upgrading a connection.

doc.go
```diff
@@ -147,9 +147,15 @@
...
+// from an HTTP connection to a websocket connection. This method performs
+// origin policy checking using the CheckOrigin field associated with the
+// Upgrader instance.
+//
+// By contrast, the deprecated package-level Upgrade function
```
garyburd: > The deprecated package-level Upgrade function does not perform origin checking. The application is responsible for checking the Origin header before calling the Upgrade function.

server.go
```diff
@@ -228,12 +228,14 @@ func (u *Upgrader) Upgrade(w http.ResponseWriter, r *http.Request, responseHeade
...
 //
-// This function is deprecated, use websocket.Upgrader instead.
+// Upgrade upgrades the HTTP server connection to the WebSocket protocol.
 //
-// The application is responsible for checking the request origin before
-// calling Upgrade. An example implementation of the same origin policy is:
+// Note that the application is responsible for checking the request origin
```
garyburd: > Upgrade does not perform origin checking. The application is responsible for checking the Origin header before calling Upgrade. An example implementation of the same origin policy check is:

doc.go
```diff
@@ -6,9 +6,9 @@
 //
 // Overview
 //
-// The Conn type represents a WebSocket connection. A server application uses
-// the Upgrade function from an Upgrader object with a HTTP request handler
-// to get a pointer to a Conn:
+// The Conn type represents a WebSocket connection. Within the context of an
```
garyburd: > The Conn type represents a WebSocket connection. A server application calls the Upgrader.Upgrade method from an HTTP request handler to get a *Conn.

doc.go
fuzzybear3965: You're right. I don't see the harm, though, in including this in the Overview section. You could stipulate that it's redundant; I'm not arguing with that. But redundancy isn't always harmful and I think, in this case, this is good information to include in the Overview. It clarifies the distinction between the Upgrader.Upgrade method and the package-level Upgrade function immediately, setting the user on the right path.

Personally, I found myself scrolling between the overview section, the documentation for the package-level Upgrade function and the Upgrader.Upgrade method to understand the difference between the two. Now, it's obvious to me what that difference is. But, having this nota bene in the Overview section would have saved me some time.

I don't expect you to find this addition that useful (as the package's primary author). But, I believe that it could prove useful to other users in the future and I don't imagine that adding it would cause universal frustration.

doc.go
fuzzybear3965: But, the key idea here is that it's the only non-deprecated mechanism for upgrading a connection. It isn't the only mechanism for upgrading a connection. Thus, due to Upgrade's deprecation we recommend Upgrader.Upgrade.

PR #284, opened 2017-09-14
add small docs for clarity argument

faisalburhanudin:


garyburd:
The argument is declared as in `int` in the function declaration. There's no need to repeat the type in comment.   

All message types are supported, not just TextMessage, BinaryMessage.

The same comment applies to `NextWriter`.

How about this:

> All message types (TextMessage, BinaryMessage, CloseMessage, PingMessage and PongMessage) are supported.

garyburd:
See f918560c4c3d6a271185320bb1d4a1e40f77297f/

PR #291, opened 2017-10-10
refine Conn.SetReadDeadline

sunlei99:
Under my circumstance

I want to use the websocket connection **Half-duplex**

e.g.
```go
    conn, _ := upgrader.Upgrade(w, r, nil)
    for {
        conn.SetReadDeadline(time.Now().Add(time.Second * 3))
        err := c.Conn.ReadJSON(&p)
        // do something...
        if err != nil && strings.Contains(err.Error(), "i/o timeout") {
            // read timeout do something
        } else if err == nil {
            // parse the data from client
        } else {
            // process the other err
        }
    }
```

**but** 
 **readErr** of Conn is always not nil after the first read timeout so the code above will panic due to 
the code in Conn.NextReader
```go
func (c *Conn) NextReader() (messageType int, r io.Reader, err error) {
        ...
        // always not nil after first timeout
	for c.readErr == nil {
		...
	}

	// panic...
	c.readErrCount++
	if c.readErrCount >= 1000 {
		panic("repeated read on failed websocket connection")
	}

	return noFrame, nil, c.readErr
}
```

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -13,10 +13,10 @@ import (
 	"math/rand"
 	"net"
 	"strconv"
+	"strings"
 	"sync"
 	"time"
 	"unicode/utf8"
-	"strings"
 )
 
 const (
--- a/conn.go
+++ b/conn.go
@@ -16,6 +16,7 @@ import (
 	"sync"
 	"time"
 	"unicode/utf8"
+	"strings"
 )
 
 const (
@@ -1029,6 +1030,9 @@ func (c *Conn) ReadMessage() (messageType int, p []byte, err error) {
 // all future reads will return an error. A zero value for t means reads will
 // not time out.
 func (c *Conn) SetReadDeadline(t time.Time) error {
+	if c.readErr != nil && strings.Contains(c.readErr.Error(), "i/o timeout") {
+		c.readErr = nil
+	}
 	return c.conn.SetReadDeadline(t)
 }
 


garyburd:
This PR does not handle the case where the timeout happens after one byte of a message is read.

Use a goroutine.

sunlei99:
Does that mean the connection is not available and should be abandoned after the timeout set by 
SetReadDeadline?  

The function SetReadDeadline should only be called ONCE?

Under the circumstance above, the only way is invoking a new goroutine?

garyburd:
The connection should be closed and abandoned after any error, including timeout.

Most applications that call SetReadDeadline with a non-zero time will call SetReadDeadline multiple times.  

Use time.Ticker time.After, goroutines, etc. to execute code at specified times.

sunlei99:
@garyburd thanks!

    About the second question: The function SetReadDeadline should only be called ONCE?

readErr of Conn is always not nil after the first read timeout, which is set by SetReadDeadline.
So does it means that 
The function SetReadDeadline should only be called ONCE?

garyburd:
If an application calls SetReadDeadline at all,, then it's typical to call the method multiple times.

sunlei99:
@garyburd Yes! I intended to call the method multiple times.
but the program will panic, because conn.readErr is always not nil 

```go
    conn, _ := upgrader.Upgrade(w, r, nil)
    for {
        conn.SetReadDeadline(time.Now().Add(time.Second * 3))
        err := conn.ReadJSON(&p)
        // do something...
        if err != nil && strings.Contains(err.Error(), "i/o timeout") {
            // read timeout do something
        } else if err == nil {
            // parse the data from client
        } else {
            // process the other err
        }
    }
```


but
readErr of Conn is always not nil after the first read timeout so the code above will panic due to
the code in Conn.NextReader
```go
func (c *Conn) NextReader() (messageType int, r io.Reader, err error) {
        ...
        // always not nil after first timeout
	for c.readErr == nil {
		...
	}

	// panic...
	c.readErrCount++
	if c.readErrCount >= 1000 {
		panic("repeated read on failed websocket connection")
	}

	return noFrame, nil, c.readErr
}
```

garyburd:
SetReadDeadline can be called multiple times to extend deadlines, but the method can only be called once for a given read operation.

Use time.Ticker time.After, goroutines, etc. to execute code at specified times.

PR #300, opened 2017-11-22
origin host check should be case insensitive

ddollar:
According to [RFC 4343](https://tools.ietf.org/html/rfc4343) DNS hostnames should be case insensitive. This modifies the `checkSameOrigin()` helper to compare hosts without case sensitivity.

### DIFF
--- a/server.go
+++ b/server.go
@@ -76,7 +76,7 @@ func checkSameOrigin(r *http.Request) bool {
 	if err != nil {
 		return false
 	}
-	return strings.ToLower(u.Host) == strings.ToLower(r.Host)
+	return equalASCIIFold(u.Host, r.Host)
 }
 
 func (u *Upgrader) selectSubprotocol(r *http.Request, responseHeader http.Header) string {
--- a/server.go
+++ b/server.go
@@ -76,7 +76,7 @@ func checkSameOrigin(r *http.Request) bool {
 	if err != nil {
 		return false
 	}
-	return strings.EqualFold(u.Host, r.Host)
+	return strings.ToLower(u.Host) == strings.ToLower(r.Host)
 }
 
 func (u *Upgrader) selectSubprotocol(r *http.Request, responseHeader http.Header) string {
--- a/server.go
+++ b/server.go
@@ -76,7 +76,7 @@ func checkSameOrigin(r *http.Request) bool {
 	if err != nil {
 		return false
 	}
-	return u.Host == r.Host
+	return strings.EqualFold(u.Host, r.Host)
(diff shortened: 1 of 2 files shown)


garyburd:
Has this issue come up in practice?

The linked RFC does not say that Unicode case-folding applies to hostname comparisons.  Please provide more background on hostname comparisons.

elithrar:
A HTTP Origin is also not a DNS hostname, to nitpick. What does the latest
72xxx HTTP (or CORS) standard say with respect to that?

It may be the same, but we should be cautious about these kind of changes.

(I’m not at a computer but will check this weekend)

ddollar:
Yes this came up in practice. AWS ELBs by default show their host names with uppercase characters. Spent a while tracking down the bad origin errror.

garyburd:
Are you saying that the ELB converts the Host header sent by the client to uppercase?

ddollar:
If you use the x/net websocket library as a client starting with a hostname that contains uppercase characters (such as provided by AWS as the hostname of an ELB) then you would end up with an unexpected Origin mismatch.

https://github.com/golang/net/blob/master/websocket/hybi.go#L419

garyburd:
re x/net/websocket:

- Because the x/net/websocket implementation of the protocol is incomplete, the x/net/websocket package does not interoperate with the Gorilla package in some scenarios. 
-  Because the Origin header is not required in typical uses of the x/net/websocket package, the package should not force the use of the header in the WebSocket handshake.

re hostname comparisons:

RFC 4343 says that ASCII case folding should be used when comparing hostnames in DNS. 

Internationalized Domain Names are encoded to ASCII at a layer above DNS. RFC 3491 (nameprep) describes how to prepare Internationalized Domain Names for comparision.

The name comparison in checkSameOrigin should use nameprep, but I don't want to add a dependency on a nameprep package.

The comparison used in checkSameOrigin should be safe with regards to nameprep: if equal(s, t) is true, then nameprep(s) == nameprep(t).

The comparison s == t is safe, but overly restrictive.

The comparison [equalASCIIFold(s, t)](https://github.com/gorilla/websocket/blob/23059f29570f0e13fca80ef6aea0f04c11daaa4d/util.go#L132) is also safe and may solve your problem.  I will accept a PR with this change.

I have not dug into nameprep far enough to determine if strings.EqualsFold(s, t) is safe. I will appreciate some help from somebody with knowledge in this domain.

ddollar:
Made that change, anything else needed here?

ddollar:
Thanks!

PR #306, opened 2017-12-01
Improve bad handshake error text

garyburd:
Change the error text for bad handshake errors from

    websocket: not a websocket handshake:

to:

    websocket: the client is not using the websocket protocol:

The new text should be more helpful to developers who do not know or
understand the details of the protocol.


elithrar:
Good move 👌🏼

PR #312, opened 2017-12-09
examples/chat/client.go: avoid allocating []byte{} for PingMessage

JordanP:
It's useless and only gives more work to the GC.


garyburd:
[]byte{} does not allocate memory.

PR #315, opened 2017-12-21
Improve emample: chat

invzhi:
- Use empty struct instead of bool in Hub struct:
`clients map[*Client]bool` -> `clients map[*Client]struct{}`
- Update README.md


elithrar:
The empty struct 'trick' is a neat optimization, but it's a micro-optimization and can be confusing to newcomers. I'd prefer that we keep the (simpler, clearer) `map[*Client]bool` type in an example/demo app.

invzhi:
I understand why. Thanks for your explanation.

garyburd:
PR not accepted for the reason stated by @elithrar.

PR #317, opened 2017-12-27
Fix mistake in chat example

invzhi:
I am not good at JavaScript, but I found this judgment is useless, everytime it will be `true`.


garyburd:
This change forces every new message to come into view.  This is annoying when scrolled back to read old messages.  From a usability perspective, the code is better as is.

invzhi:
Sorry, I misunderstand this code. Thanks for your explanaton. You can close this PR now.

Issue #316, opened 2017-12-26
context.Context support


hanbang-wang:
Support wanted for better coroutine manipulation.

garyburd:
Describe the problem you are trying to solve in more detail.

hanbang-wang:
`context` package, as declared in standard packages and document available here [https://golang.org/pkg/context/](https://golang.org/pkg/context/), is used to handle network requests and responses in goroutine. For example, if a connection holds a cancellable context, does the standard `http` package, and the `context` is canceled, the whole request will be canceled, including connecting, reading or writing, and the open socket will be closed. 

Honestly, this is pretty self-explanatory if you know what `context` package does. Furthermore, the `Dial` param in `http.Transport` is already deprecated, and the `DialContext` is a great replacement for that, so I believe it should be the same for this `websocket` package as well.

garyburd:
I am asking you to describe the problem you are trying to solve in your application, not a description of the context package.

hanbang-wang:
For example, when I am dialing, instead of `HandshakeTimeout` I want to use `context` to `cancel` it.

garyburd:
Why does the timeout not work for your application?  Describe the problem.

I am not going to sprinkle context dust over the package with the hope that it will magically solve actual problems.

hanbang-wang:
As I said, context is created and suggested for problems like this. Let me give you a more detailed and *realistic* situation.


There's a MITM websocket service running, two goroutines respectively handle connections from (client to self) and (self to the real server). 

Say that I am reading data from (self to the real server) connection and to send it to the client side, but since the real server is not sending anything, the procedure is blocked. Now, the client closed the (client to self) connection, so I am supposed to be closing (self to the real server) connection as well. However, since the (self to the real server) goroutine is still blocked on reading, I can do nothing about it. 

Now I know that I can set a ReadTimeout on (self to the real server) side, and detect if the other (client to self) connection is still alive after every timeout, but a cancellable context would be a perfect and quick solution for this: simply calls `cancel()` and the reading is not blocked anymore and returns an error instead.

garyburd:
Context cancelation is **not** the right solution for handling close in a websocket proxy. Let me know if this is an actual problem you are encountering. If it is, then I'lll take the time to describe how to handle close in a proxy.

I know what the context package is and the problems the package is intended to solve.  I am willing to add context related features based on actual application requirements. I don't want to spend time implementing or reviewing features that nobody needs.

[...]

PR #322, opened 2018-01-04
Expose internal buffer reader for read raw data

Juliens:
We use gorilla/websocket in [traefik](https://github.com/containous/traefik) in order to reverse-proxying websocket traffic.

What we want to do is to use gorilla/websocket for the Upgrade processes (client <=> reverse-proxy and reverse-proxy <=> server proxyfied) and then `io.Copy` the underlying connection in order to transfer the raw traffic. (like [here](https://github.com/vulcand/oxy/blob/master/forward/fwd.go#L346)  )

The issue we encounter with this, is that some servers embed websockets messages in the Switching Protocol body.
If it happens, those messages are not in the Underlying connection and are in the buffer reader in websocket.Conn ( `br` attribute).

That's why, we need to be able to access this buffer reader.

WDYT ?

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -1119,6 +1119,12 @@ func (c *Conn) UnderlyingConn() net.Conn {
 	return c.conn
 }
 
+// BufferReader returns the internal buffer reader. This can be used to read
+// the raw data after upgrade success
+func (c *Conn) BufferReader() *bufio.Reader {
+	return c.br
+}
+
 // EnableWriteCompression enables and disables write compression of
 // subsequent text and binary messages. This function is a noop if
 // compression was not negotiated with the peer.


garyburd:
Explain in more detail on how messages end up i the buffered reader.  Does the application process some messages normally before io.Copying the underlying connection?

Juliens:
We don't need to read message before io.Copying the underlying connection.

I have tried naively to just read the first message with ReadMessage, rewrite it to the client, and then start io.Copying, but after deeper analyse, I understand that it was not enough because the server can send multiple messages in the body of the switching protocol response.

What we really need today is only:
 - Do the Upgrade with client
 - Do the Dial with the server
 - io.Copying underlying connection

garyburd:
I read the issue to say that you need this feature for a connection returned from Upgrade.  Are you saying that you need this for a Dialed connection?

Juliens:
I think code tell more than word 😛 

What I want to do
```golang
underlyingConn, err := websocket.Upgrader{}.Upgrade(w, req, resp.Header)
targetConn, resp, err := websocket.DefaultDialer.Dial(server, outReq.Header)

io.Copy(underlyingConn.UnderlyingConn(), targetConn.BufferReader())
io.Copy(targetConn.UnderlyingConn(), underlyingConn.UnderlyingConn())
```

garyburd:
The code above assumes that the protocol negotiated between the client and proxy in Upgrade is identical to the protocol negotiated between the proxy and server in Dial. That's not a guarantee that I want to make.

The application should either copy messages or it should handle the websocket handshake directly. 

Separate from the protocol negotiation and access to data read from the dialed connection, the code does not work as a general purpose websocket proxy.

Juliens:
I'm really curious about your feedback and I would have liked to talk about it with you in a less formal way.
Do you think you can come on our [slack](https://traefik.herokuapp.com/) ?
Or another way to talk about it ?

garyburd:
I went to you slack. Now what?

The problem with the code is that it does not allow the client and server to negotiate a subprotocol, nor does it pass headers back and forth.  The logic should look something like this:
         
         dial server using information from request headers
         if server does not upgrade the connection then
             respond to client's http request using response from server
             return
         upgrade connection to client using headers from dial response
         start copying data

The proxy can copy raw data as in the the above or copy messages as in [this comment](https://github.com/gorilla/websocket/issues/316#issuecomment-353987393).   The advantage of copying messages is that proxy can inject pings in the stream to detect dead peers.  The advantage  of copying raw data is that there's less overhead on the proxy.

If the proxy does copy the raw data, then the handshake should be between the client and server.  The proxy passes the handshake through with simple modifications such as adding the x-forwarded-for header. In this scenario, it does not make sense to use the Dial and Upgrade functionality in this package.

garyburd:
Closing because the feature is not needed for a correct program.

PR #326, opened 2018-01-10
Replace "frame" with "message" in documentation

garyburd:
The documentation sometimes used the term "frame" when referring to
single frame control messages.  Use the term "message" for consistency
in the documentation and to hide a detail that most application
programmers do not need to know about.


elithrar:
👌🏼

PR #333, opened 2018-01-25
Improve check origin documentation

garyburd:
Remove the example code to disable origin checks from the documentation.
I am concerned that developers are copying the code without
understanding the security implications of the code. Most applications
should not use this code.

Change the bad origin error message to mention Upgrader.CheckOrigin

### DIFF
--- a/doc.go
+++ b/doc.go
@@ -144,15 +144,8 @@
 // method fails the WebSocket handshake with HTTP status 403.
 //
 // If the CheckOrigin field is nil, then the Upgrader uses a safe default: fail
-// the handshake if the Origin request header is present and not equal to the
-// Host request header.
-//
-// An application can allow connections from any origin by specifying a
-// function that always returns true:
-//
-//  var upgrader = websocket.Upgrader{
-//      CheckOrigin: func(r *http.Request) bool { return true },
-//  }
+// the handshake if the Origin request header is present and the Origin host is
+// not equal to the Host request header.
 //
 // The deprecated package-level Upgrade function does not perform origin
 // checking. The application is responsible for checking the Origin header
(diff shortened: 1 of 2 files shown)


doc.go
```diff
@@ -147,13 +147,6 @@
 // the handshake if the Origin request header is present and not equal to the
 // Host request header.
 //
-// An application can allow connections from any origin by specifying a
-// function that always returns true:
-//
-//  var upgrader = websocket.Upgrader{
-//      CheckOrigin: func(r *http.Request) bool { return true },
```
elithrar: Add additional context to the `CheckOrigin` field in the `Upgrader` struct? https://godoc.org/github.com/gorilla/websocket#Upgrader

from:
```
    // CheckOrigin returns true if the request Origin header is acceptable. If
    // CheckOrigin is nil, the host in the Origin header must not be set or
    // must match the host of the request.
```

to:
```
    // CheckOrigin returns true if the request Origin header is acceptable. If
    // CheckOrigin is nil, the host in the Origin header must not be set or
    // must match the host of the request. You should carefully valid
    // Origin headers, and not blindly accept requests from any origin.
```

garyburd: @elithrar Please comment on my most recent change.

server.go
```diff
@@ -44,8 +44,12 @@ type Upgrader struct {
 	Error func(w http.ResponseWriter, r *http.Request, status int, reason error)
 
 	// CheckOrigin returns true if the request Origin header is acceptable. If
-	// CheckOrigin is nil, the host in the Origin header must not be set or
-	// must match the host of the request.
+	// CheckOrigin is nil, then a safe default is used: return false if the
```
elithrar: Perfect.

Issue #339, opened 2018-01-31
Want to implement rfc7692, but writer side can not be implemented


smith-30:
Hi,

I'd like to use the context-takeover mechanism defined in rfc7692.
I forked and developed it and I could implement the reader side.
This [branch] (https://github.com/smith-30/websocket/tree/feature/upgrade_writer) is the newest.

However, I am in trouble because I can not implement the writer.

__Implementation I'm thinking__

Attempting to implement context-takeover by attaching flateWriteWrapper to Conn struct.

In flateWriteWrapper, attach flat.Writer called with flat.NewWriterDict.
https://github.com/smith-30/websocket/blob/ee46f8548a106a02264f711a1838887fd3cf58cf/conn.go#L518-L536
I do not want to make flateWriter every time I make a call.
Because performance is very poor.
Not using Pool is because GC may clean it without permission.
Avoid the window of flateWriter disappearing and inconsistency with reader side.

However, in my implementation I can not initialize the truncWriter passed to flateWriter.
In the second execution, truncWriter has state and fails in compress processing.
I'd like to reset the state of truncWriter after compression
Is there a good way to do it?

I am sorry for my poor English.
It would be extremely helpful If you review my implementation..

smith-30:
__note__

The client of context-takeover uses python's one.
https://github.com/aaugustin/websockets

I will post code that I am using for debugging if necessary.

garyburd:
The code in conn.go is written to support context takeover.  When context takeover is used, the compression factory functions (newDecompressionReader, newCompressionWriter) maintain per-connection state through a closure or a [method value](https://golang.org/ref/spec#Method_values).  This state will include the flate.Writer or flate.Reader for the connection. A method value implementation will look something like this:

    type struct contextTakeoverWriterFactory {
         flateWriter flate.Writer
         // and other fields
    }

    func (f *contextTakeoverWriterFactory) newCompressionWriter(w io.WriteCloser, level int) io.WriteCloser  {
        // wedge w under f.flateWriter
        // return wrapper around f.flateWriter
    }

The Upgrader.Upgrade and DIaler.Dial methods will setup the factory and assign a method value to the factory function fields:

      var f contextTakeoverWriterFactory
      f.flateWriter = ... // create flate.Writer for this conneciton
     // Set other fields
      c.newCompressionWriter = f.newCompressionWriter

[...]

PR #342, opened 2018-02-07
rfc7692 context-takeover implemention

smith-30:
related #339


client_server_test.go
```diff
@@ -42,6 +42,8 @@ var cstDialer = Dialer{
 
 type cstHandler struct{ *testing.T }
 
+type cstContextTakeoverHandler struct{ *testing.T }
```
garyburd: Use `cstHandler` and `cstServer` for all tests.

- Add field to `cstHandler` to specify context takeover options.  
- Add arguments to `newServer` to pass options through to the handler.
- Change `cstHandler.ServeHTTP` to send and receive two messages. 
- Modify `sendRecv` to send and receive two messages.  Use loop over messages instead of duplicating gobs of code.

      for _, message := range []string{"Hello World", "Can you read message?"} {
            ...
      }

conn.go
```diff
@@ -81,20 +83,14 @@ const (
...
-	timeout   bool
-}
+type (
+	// netError satisfies the net Error interface.
+	netError struct {
+		msg       string
+		temporary bool
+		timeout   bool
+	}
+)
```
garyburd: Do not use type ( ) for a single type.  Restore to original code.

conn.go
```diff
@@ -178,6 +174,14 @@ var (
 	errBadWriteOpCode      = errors.New("websocket: bad write message type")
 	errWriteClosed         = errors.New("websocket: write closed")
 	errInvalidControlFrame = errors.New("websocket: invalid control frame")
+
+	// ErrCloseSent is returned when the application writes a message to the
+	// connection after sending a close message.
+	ErrCloseSent = errors.New("websocket: close sent")
```
garyburd: Move ErrCloseSent and ErrReadLimit back to where they were. There's are a lot of changes to review. It's easier without changes unrelated to the feature.

conn.go
```diff
@@ -259,8 +263,10 @@ type Conn struct {
 	readErrCount  int
 	messageReader *messageReader // the current low-level reader
 
-	readDecompress         bool // whether last read frame had RSV1 set
-	newDecompressionReader func(io.Reader) io.ReadCloser
+	readDecompress         bool                          // whether last read frame had RSV1 set
+	newDecompressionReader func(io.Reader) io.ReadCloser // arges may flateReadWrapper struct
+
+	contextTakeover bool
```
garyburd: This field should not be required. All  connection logic for compression should be pushed into the the decompression reader and decompression writer.

smith-30:
@garyburd Sorry for the delay after reviewing it.  
I fixed the part you pointed out.

smith-30:
hi, Is there progress?
 I'm happy if you give feedback. 😃

garyburd:
It will be another few days before I have time to review this further. There's a lot to review here. 

Please test the feature with Autobahn.  See https://github.com/gorilla/websocket/tree/master/examples/autobahn.

smith-30:
@garyburd thanks!
I will try to pass the test at Autobahn.

smith-30:
@garyburd 
I did a test but there was no new error.
However, in both branches case 12.4.5 ~ 6, 12.4.8 ~ 11, 12.4.13 ~ 12.4.18 have failed.
Can we ignore this error?

PR branch's result
https://smith-30.github.io/websocket/autobahn/context-takeover/index.html

master branch's result
https://smith-30.github.io/websocket/autobahn/master/index.html

garyburd:
The tests 12.4.5 ~ 6, 12.4.8 ~ 11, 12.4.13 ~ 12.4.18 fail because the test times out.  Ignore for now.

Did you edit to the test application to enable context takeover?

smith-30:
> The tests 12.4.5 ~ 6, 12.4.8 ~ 11, 12.4.13 ~ 12.4.18 fail because the test times out. I opened an issue for this.

thanks!

> Did you edit to the test application to enable context takeover?

sorry, I didn't add context-takeover option.
I added it and re-run. It seems that all the tests are passing.

server.go
https://github.com/gorilla/websocket/pull/342/commits/658a2fb894f477e0c12f4b2e8be94a7c1111736b

result
https://smith-30.github.io/websocket/autobahn/context-takeover/index.html

conn.go
```diff
@@ -259,8 +261,8 @@ type Conn struct {
 	readErrCount  int
 	messageReader *messageReader // the current low-level reader
 
-	readDecompress         bool // whether last read frame had RSV1 set
-	newDecompressionReader func(io.Reader) io.ReadCloser
+	readDecompress         bool                          // whether last read frame had RSV1 set
+	newDecompressionReader func(io.Reader) io.ReadCloser // arges may flateReadWrapper struct
```
JordanP: I don't understand the code comment here. What "arges may flateReadWrapper struct" mean ?

conn.go
```diff
@@ -749,9 +752,10 @@ func (c *Conn) WriteMessage(messageType int, data []byte) error {
 	if err != nil {
 		return err
 	}
-	if _, err = w.Write(data); err != nil {
+	if _, err := w.Write(data); err != nil {
```
JordanP: Why this change ?

conn.go
smith-30: sorry, this comment is unnecessary.
This was left for myself during initial development.

thanks.

conn.go
smith-30: Three months ago so I do not remember why I made this change.
There is no need to reassign here. I reverted to the original code.

Thanks.

smith-30: Also, I returned those which were unnecessary changes such as line feed.

JordanP: Thanks good job. I am really interested  in the PR, I may try it on part of our production system next week or the week after.

smith-30: I would like you to point out any improvement points you tried as a result.
I hope that good results will come!

JordanP: TBH I didn't test it yet. It's still in my todo-list but I have a lot to do for the moment, I am sorry.

smith-30:
I prepared a branch that organized commit so I will replace it if necessary.
[this](https://github.com/smith-30/websocket/commits/feature/rebase_context-takeover) branch is created by 

```console
$ git merge --squash feature/context-takeover
```

JordanP:
Could you make a benchmark with and without CompressionContextTakeOver and compare the memory allocation/memory usage. I ran some tests internally and I noticed that memory usage was multiplied by 10. So I am not sure, maybe there's something wrong with my bench or a memory leak in the code...

doc.go
```diff
@@ -171,10 +171,17 @@
 //
 //  conn.EnableWriteCompression(false)
 //
-// Currently this package does not support compression with "context takeover".
+// Currently this package support compression with "context takeover".
```
theckman: Pluralize `support`.

theckman:
@smith-30 have you had the opportunity to look at the comment from @JordanP around the size of magnitude of the memory increase?

smith-30:
@theckman  I'm sorry, I missed replying and I was late. thank you for your comment.

First of all, @JordanP pointed out an increase in memory usage, but since it has not happened with my product, I do not know if I write test code.
By the way, the benchmark environment I tried is 3 for api server and 180 for client. This is because it is a very high transmission frequency that one client sends.

I think that it will take 1-2 weeks for that response.

After that, I will also modify the comments. Thank you for review.

JordanP:
On my side, my workload is different: tens of thousand clients, each sending one Websocket message every 5sec (i.e not frequently).

smith-30:
So, in this implementation, I think that it has a maximum dictionary of 32 kb to share a dictionary between client and server. Therefore, using context - takeover will increase memory usage.

doc.go
smith-30: thanks!

hexdigest:
Hi everyone!

Is there a chance to see this PR in master?
Benchmarks show no performance degradation:

Master:
```
goos: linux
goarch: amd64
BenchmarkWriteNoCompression-4     	 5000000	       331 ns/op	      48 B/op	       1 allocs/op
BenchmarkWriteWithCompression-4   	  300000	      4942 ns/op	     164 B/op	       3 allocs/op
BenchmarkBroadcast/NoCompression-4         	     200	   9503267 ns/op	       0 B/op	       0 allocs/op
BenchmarkBroadcast/WithCompression-4       	      50	  30607582 ns/op	 1602198 B/op	   30010 allocs/op
BenchmarkBroadcast/NoCompressionPrepared-4 	     200	  10139983 ns/op	     167 B/op	       1 allocs/op
BenchmarkBroadcast/WithCompressionPrepared-4         	     100	  10455902 ns/op	   12451 B/op	       3 allocs/op
```

PR:
```
BenchmarkWriteNoCompression-4                      	 5000000	       325 ns/op	      48 B/op	       1 allocs/op
BenchmarkWriteWithCompression-4                    	  300000	      4921 ns/op	     164 B/op	       3 allocs/op
BenchmarkWriteWithCompressionOfContextTakeover-4   	  300000	      4806 ns/op	      60 B/op	       2 allocs/op
BenchmarkBroadcast/NoCompression-4                 	     200	   9663945 ns/op	       0 B/op	       0 allocs/op
BenchmarkBroadcast/WithCompression-4               	      50	  29305592 ns/op	 1529686 B/op	   30006 allocs/op
BenchmarkBroadcast/NoCompressionPrepared-4         	     200	  10326534 ns/op	     166 B/op	       1 allocs/op
BenchmarkBroadcast/WithCompressionPrepared-4       	     100	  10340785 ns/op	   12396 B/op	       3 allocs/op
```

Since this feature is optional why not have it in master?

JordanP:
@hexdigest great that you have interest in this too. Could you share some numbers about your workload ? With compression enabled, how much more memory does your app need ? What about the CPU cost, did you measure some impact with/without this ?

hexdigest:
@JordanP I don't think these are legit questions. These two modes: with context takeover and with no context takeover serve different purposes. In my case, there are a few thousand long living clients that send ~1 message per second where every message is more or less the same small payload (around 200 bytes) with very few differences. For my case, having context takeover is crucial because we're paying for the traffic.

I understand that the current compression implementation in the master branch uses the pool to minimize the overhead of memory allocations. This approach is especially good for the applications with hundreds of thousands of short living connections where each client rarely sends big chunks of random data. Unfortunately this is not my and @smith-30 use.

Per message CPU costs are shown here:
```
BenchmarkWriteWithCompression-4                    	  300000	      4921 ns/op	     164 B/op	       3 allocs/op
BenchmarkWriteWithCompressionOfContextTakeover-4   	  300000	      4806 ns/op	      60 B/op	       2 allocs/op
```
With same compression level context takeover has less impact on CPU per message because there are less allocations.
Of course there are hidden costs for allocating/freeing memory for the flate reader/writer for each connection because the pool can't be used in the case of context takeover mode. But we're ready to pay this price in exchange for 10x-20х better compression. Considering the fact that this feature is optional and disabled by default why not?

verebes:
@smith-30 & @elithrar, what is the reason this PR has not yet been merged?

smith-30:
@verebes I think it's because there is no maintainer to make the merge decision..
https://github.com/gorilla/websocket/issues/370

PR #356, opened 2018-03-04
Add a default handshake timeout of 5 seconds

pcarranza:
I'm adding this because while working with `nlopes/slack`, which in turn uses gorilla to deal with WebSockets, in a weird MTU related network failure, the behavior was that the client locks forever because there is no handshake timeout by default.

This way Gorilla will simply err out when the 5 seconds handshake timeout is passed (which, IMO, 5 seconds is a lot of time for a handshake)

### DIFF
--- a/client.go
+++ b/client.go
@@ -41,7 +41,7 @@ func NewClient(netConn net.Conn, u *url.URL, requestHeader http.Header, readBufS
 		NetDial: func(net, addr string) (net.Conn, error) {
 			return netConn, nil
 		},
-		HandshakeTimeout: 5 * time.Second,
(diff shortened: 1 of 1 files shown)


garyburd:
The timeout must be conservative. Applications that work without the timeout should continue to work with this change.

The timeout is for DNS lookup, TCP handshake, proxy negotiation, TLS handshake, and the actual WebSocket handshake.   Five seconds is too short. Let's go with 45 seconds.

pcarranza:
Thanks @garyburd, I'll make the change later

garyburd:
See 8c40c0b5bde0d67e79aeda4274a303ebcb4cebc3

PR #383, opened 2018-05-22
Add context in the Dialer

Juliens:
In order to enable httptrace, you need to add a context to the request.

With this modification, you can add the context to the request if you give the context to the new DialWithContext.

### DIFF
--- a/client.go
+++ b/client.go
@@ -6,6 +6,7 @@ package websocket
 
 import (
 	"bytes"
+	"context"
 	"crypto/tls"
 	"errors"
 	"io"
@@ -51,6 +52,10 @@ type Dialer struct {
 	// NetDial is nil, net.Dial is used.
 	NetDial func(network, addr string) (net.Conn, error)
 
+	// NetDialContext specifies the dial function for creating TCP connections. If
+	// NetDialContext is nil, net.DialContext is used.
+	NetDialContext func(ctx context.Context, network, addr string) (net.Conn, error)
+
 	// Proxy specifies a function to return a proxy for a given
 	// Request. If the function returns a non-nil error, the
 	// request is aborted with the provided error.
@@ -84,6 +89,19 @@ type Dialer struct {
 	Jar http.CookieJar
 }
 
+// Dial creates a new client connection. Use requestHeader to specify the
+// origin (Origin), subprotocols (Sec-WebSocket-Protocol) and cookies (Cookie).
+// Use the response.Header to get the selected subprotocol
+// (Sec-WebSocket-Protocol) and cookies (Set-Cookie).
+//
+// If the WebSocket handshake fails, ErrBadHandshake is returned along with a
+// non-nil *http.Response so that callers can handle redirects, authentication,
+// etcetera. The response body may not contain the entire response and does not
+// need to be closed by the application.
+func (d *Dialer) Dial(urlStr string, requestHeader http.Header) (*Conn, *http.Response, error) {
+	return d.DialWithContext(urlStr, requestHeader, context.Background())
+}
+
 var errMalformedURL = errors.New("malformed ws or wss URL")
 
 func hostPortNoPort(u *url.URL) (hostPort, hostNoPort string) {
@@ -113,17 +131,18 @@ var DefaultDialer = &Dialer{
 // nilDialer is dialer to use when receiver is nil.
 var nilDialer Dialer = *DefaultDialer
 
-// Dial creates a new client connection. Use requestHeader to specify the
+// DialWithContext creates a new client connection. Use requestHeader to specify the
 // origin (Origin), subprotocols (Sec-WebSocket-Protocol) and cookies (Cookie).
 // Use the response.Header to get the selected subprotocol
 // (Sec-WebSocket-Protocol) and cookies (Set-Cookie).
+//
+// The context will be used in the request and in the Dialer
 //
 // If the WebSocket handshake fails, ErrBadHandshake is returned along with a
 // non-nil *http.Response so that callers can handle redirects, authentication,
 // etcetera. The response body may not contain the entire response and does not
 // need to be closed by the application.
-func (d *Dialer) Dial(urlStr string, requestHeader http.Header) (*Conn, *http.Response, error) {
-
+func (d *Dialer) DialWithContext(urlStr string, requestHeader http.Header, ctx context.Context) (*Conn, *http.Response, error) {
 	if d == nil {
 		d = &nilDialer
 	}
@@ -161,6 +180,7 @@ func (d *Dialer) Dial(urlStr string, requestHeader http.Header) (*Conn, *http.Re
 		Header:     make(http.Header),
 		Host:       u.Host,
 	}
+	req = req.WithContext(ctx)
 
 	// Set the cookies present in the cookie jar of the dialer
 	if d.Jar != nil {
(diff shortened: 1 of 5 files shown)


dottyjones:
If a context argument is added, then most people will expect the core functionality of cancelation and deadlines to work.   The context should be [plumbed in here](https://github.com/gorilla/websocket/blob/21ab95fa12b9bdd8fecf5fa3586aad941cc98785/client.go#L213-L217).  Perhaps a Dialer.NetDialContext field is also required.

Juliens:
@dottyjones I agree that most people will expect the core functionnality of context.

But before going deeper in the change, I realise that this changes may need to depreciate golang before 1.7 support (or handling compatibility, but this may need a lots of changes).

In addtion, after finding #316, @garyburd do you think it's a real use case for adding context?

garyburd:
Tracing is a valid use case for adding DialWithContext. If this method is added, then it should implement cancelation.

elithrar:
My opinion: 1.7+ only, for reasons of simplicity & age.

Juliens:
Oky, I push a new version and remove go version previous to 1.7

client.go
```diff
@@ -212,8 +232,16 @@ func (d *Dialer) Dial(urlStr string, requestHeader http.Header) (*Conn, *http.Re
...
+		if d.NetDialContext == nil {
+			netDialer := &net.Dialer{Deadline: deadline}
+			netDial = func(network, addr string) (net.Conn, error) {
+				return netDialer.DialContext(ctx, network, addr)
+			}
+		} else {
+			netDial = func(network, addr string) (net.Conn, error) {
+				return d.NetDialContext(ctx, network, addr)
+			}
+		}
```
garyburd: Flatten the nested logic.

     var netDial func(network, addr string) (net.Conn, error)
     if d.NetDialContext != nil {
          netDial = func( ...
     } else if d.NetDial != nil {
           netDial = func( ...
      } else {
	    netDialer := &net.Dialer{Deadline: deadline}
	    netDial = netDialer.Dial
      }

client.go
```diff
@@ -212,8 +232,16 @@ func (d *Dialer) Dial(urlStr string, requestHeader http.Header) (*Conn, *http.Re
 	// Get network dial function.
```
garyburd: Set `deadline` to minimum of ctx.Deadline() and time computed from d.HandshakeTimeout.

client_server_test.go
```diff
@@ -659,3 +661,31 @@ func TestSocksProxyDial(t *testing.T) {
 	defer ws.Close()
 	sendRecv(t, ws)
 }
+
+func TestDialWithContext(t *testing.T) {
+	s := newServer(t)
+	defer s.Close()
+
+	var headersWrote, requestWrote bool
+	trace := &httptrace.ClientTrace{
```
garyburd: Test for all relevant events and implement missing events as needed.  Events that seem relevant are GetConn, GotConn, ConnectDone, TLSHandshakeStart, TLSHandshakeDone,  WroteHeaders, WroteRequest, GotFirstResponseByte

client_server_test.go
```diff
@@ -659,3 +661,31 @@ func TestSocksProxyDial(t *testing.T) {
 	defer ws.Close()
 	sendRecv(t, ws)
 }
```
garyburd: Add test similar to TestHandshakeTimeout, but set deadline through context instead of Dialer.HandshakeTimeout.

client.go
```diff
@@ -84,6 +89,19 @@ type Dialer struct {
 	Jar http.CookieJar
 }
 
+// Dial creates a new client connection. Use requestHeader to specify the
```
garyburd: Eliminate duplicate documentation:

    Dial creates a new client connection by calling DialContext with a background
    context.

client.go
```diff
@@ -113,17 +131,18 @@ var DefaultDialer = &Dialer{
 // nilDialer is dialer to use when receiver is nil.
 var nilDialer Dialer = *DefaultDialer
 
-// Dial creates a new client connection. Use requestHeader to specify the
+// DialWithContext creates a new client connection. Use requestHeader to specify the
```
garyburd: Name the function DialContext.

client.go
```diff
@@ -113,17 +131,18 @@ var DefaultDialer = &Dialer{
...
 // origin (Origin), subprotocols (Sec-WebSocket-Protocol) and cookies (Cookie).
 // Use the response.Header to get the selected subprotocol
 // (Sec-WebSocket-Protocol) and cookies (Set-Cookie).
 //
+// The context will be used in the request and in the Dialer
```
garyburd: Also state that context must not be nil.

garyburd:
Instead of computing the minimum deadline from the context and handshake timeout, a new context should be created with the handshake timeout:

 	if d.HandshakeTimeout != 0 {	 
             ctx, cancel = context.WithTimeout(d.HandshakeTimeout)
             defer cancel()
       }

I bungled the flattening of the net dial computation.  Here's a better version:

      var netDial func(network, addr string) (net.Conn, error)
      if d.NetDialContext != nil {
           netDial = func(network, addr string) (net.Conn, error) { return d.NetDialContext(ctx, network, addr) }
      } else if d.NetDial != nil {
           netDial = d.NetDial
      } else {
         netDialer := &net.Dialer{Timeout: d.HandshakeTimeout}
         netDial = func(network, addr string) (net.Conn, error) { return netDialer.DialContext(ctx, network, addr) }
      }

Code that uses the local variable `deadline` should get the deadline from `ctx`.

client.go
theckman: I think it should remain as `DialWithContext`, using the standard library as inspiration: https://golang.org/pkg/net/http/#Request.WithContext

client.go
theckman: Alternatively, we could add:

```Go
if ctx == nil {
  ctx = context.Background()
}
```

theckman:
@garyburd I assume the change in the logic was intentional to make use of the `d.NetDialContext` before the `d.NetDial` one so that the context can be used?

client_server_test.go
Juliens: TLSHandshakeStart and TLSHandshakeDdone are available only since 1.8. Do you want me to handle it with conditionnal build?

client.go
```diff
@@ -113,16 +124,21 @@ var DefaultDialer = &Dialer{
...
+//
 // If the WebSocket handshake fails, ErrBadHandshake is returned along with a
 // non-nil *http.Response so that callers can handle redirects, authentication,
 // etcetera. The response body may not contain the entire response and does not
 // need to be closed by the application.
-func (d *Dialer) Dial(urlStr string, requestHeader http.Header) (*Conn, *http.Response, error) {
+func (d *Dialer) DialContext(urlStr string, requestHeader http.Header, ctx context.Context) (*Conn, *http.Response, error) {
+	if ctx == nil {
```
garyburd: Follow the convention in the standard library: Either panic when ctx == nil or continue on and let code downstream panic.

adamdecaf:
https://github.com/gorilla/websocket/pull/411 ran into old Go versions not supporting methods too. Can we drop Go 1.4 through 1.6 in CI in a separate PR?

elithrar:
I’m OK with this. 1.11 is almost out & 1.7 to 1.11 is ~2.5 years of
releases.

We can backport specific bug fixes on request (only) if they are critical.

garyburd:
Changes needed: 
-  Version specific shim functions for calling TLSHandshakeStart and TLSHandshakeStop (see the Travis log).
- Remove code to set ctx to ctx.Background() when ctx == nil. If an application wants the background context, then the application should explicitly pass ctx.Background() to DialContext.

Otherwise, this LGTM.

Issue #384, opened 2018-05-27
Fix handshake timeout 


f4nff:
client.go

```
// Copyright 2013 The Gorilla WebSocket Authors. All rights reserved.
// Use of this source code is governed by a BSD-style
// license that can be found in the LICENSE file.

package websocket

import (
	"bytes"
	"crypto/tls"
	"errors"
	"io"
	"io/ioutil"
	"net"
	"net/http"
	"net/url"
	"strings"
	"time"
	"fmt"
)

// ErrBadHandshake is returned when the server response to opening handshake is
// invalid.
var ErrBadHandshake = errors.New("websocket: bad handshake")

var errInvalidCompression = errors.New("websocket: invalid compression negotiation")

// NewClient creates a new client connection using the given net connection.
// The URL u specifies the host and request URI. Use requestHeader to specify
// the origin (Origin), subprotocols (Sec-WebSocket-Protocol) and cookies
// (Cookie). Use the response.Header to get the selected subprotocol
// (Sec-WebSocket-Protocol) and cookies (Set-Cookie).
//
// If the WebSocket handshake fails, ErrBadHandshake is returned along with a
// non-nil *http.Response so that callers can handle redirects, authentication,
// etc.
//
// Deprecated: Use Dialer instead.
func NewClient(netConn net.Conn, u *url.URL, requestHeader http.Header, readBufSize, writeBufSize int) (c *Conn, response *http.Response, err error) {
	d := Dialer{
		ReadBufferSize:  readBufSize,
		WriteBufferSize: writeBufSize,
		NetDial: func(net, addr string) (net.Conn, error) {
			return netConn, nil
		},
	}
	return d.Dial(u.String(), requestHeader)
}

// A Dialer contains options for connecting to WebSocket server.
type Dialer struct {
	// NetDial specifies the dial function for creating TCP connections. If
	// NetDial is nil, net.Dial is used.
	NetDial func(network, addr string) (net.Conn, error)

	// Proxy specifies a function to return a proxy for a given
	// Request. If the function returns a non-nil error, the
	// request is aborted with the provided error.
	// If Proxy is nil or returns a nil *URL, no proxy is used.
	Proxy func(*http.Request) (*url.URL, error)

	// TLSClientConfig specifies the TLS configuration to use with tls.Client.
	// If nil, the default configuration is used.
	TLSClientConfig *tls.Config

	// HandshakeTimeout specifies the duration for the handshake to complete.
	HandshakeTimeout time.Duration

[...]

PR #385, opened 2018-05-28
Add test for handshake deadline

dottyjones:
When looking at #384, I didn't see test to confirm handshake read and write are under a deadline.

### DIFF
--- a/client_server_test.go
+++ b/client_server_test.go
@@ -344,6 +344,63 @@ func TestDialTimeout(t *testing.T) {
 	}
 }
 
+// netConnDeadlineObserver fails test if read or write called without deadline.
+type netConnDeadlineObserver struct {
+	t           *testing.T
+	c           net.Conn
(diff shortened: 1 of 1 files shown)


client_server_test.go
```diff
@@ -344,6 +344,63 @@ func TestDialTimeout(t *testing.T) {
 	}
 }
 
+// netConnDeadlineObserver fails test if read or write called without deadline.
+type netConnDeadlineObserver struct {
+	t           *testing.T
+	c           net.Conn
+	read, write bool
```
kisielk: My only qualm is these variable names are not descriptive of their purpose, making it a bit confusing to understand how this test works. Maybe rename them to `readDeadlineUsed` and `writeDeadlineUsed` ?

dottyjones:
Please take a look again.

kisielk:
Thanks

PR #386, opened 2018-05-29
update allow bind local ip to reach network

forthxu:


dottyjones:
An application can specify the local address using the [Dialer.NetDial](https://godoc.org/github.com/gorilla/websocket#Dialer.NetDial) function:

      d := websocket.Dialer{
            NetDial: (&net.Dialer{LocalAddr: addr}).Dial
      }

What is the rationale for adding a second way to specify the local address and how does this rationale apply to other [net.Dialer](https://godoc.org/net#Dialer) fields?  Will the websocket.Dialer grow to include all net.Dialer fields?

If the LocalAddr field is added to websocket.Dialer, then the deadline should be set in a single code path:

	netDialer := &net.Dialer{Deadline: deadline}
	if len(d.LocalAddr) > 0 {
		localAddr, err := net.ResolveIPAddr("ip", d.LocalAddr)
		if err != nil {
			return nil, nil, err
		}
		netDialer.LocalAddr = localAddr
	}

garyburd:
Use the NetDial function as suggested in a previous comment.

PR #401, opened 2018-07-24
Implement subprotocol selection according to RFC6455 section 4.2.2

jjakob:
…(TODO: update documentation)

Use either Upgrader.Subprotocols or requestHeader to specify subprotocols the server wants. 
If neither is defined, the server only accepts clients with empty subprotocols.
If both are defined, requestHeader overrides Subprotocols. This can be used to e.g. manually select which subprotocol the server wants for this specific connection when calling Upgrader.Upgrade.
If multiple matching subprotocols are found, the first one is selected - server precedence.
Works fine here. Please test and send me any issues you may find.

### DIFF
--- a/server.go
+++ b/server.go
@@ -83,20 +83,53 @@ func checkSameOrigin(r *http.Request) bool {
 	return equalASCIIFold(u.Host, r.Host)
 }
 
-func (u *Upgrader) selectSubprotocol(r *http.Request, responseHeader http.Header) string {
-	if u.Subprotocols != nil {
-		clientProtocols := Subprotocols(r)
-		for _, serverProtocol := range u.Subprotocols {
-			for _, clientProtocol := range clientProtocols {
-				if clientProtocol == serverProtocol {
-					return clientProtocol
-				}
+// firstMatching returns the first matching element present in both slices and true/false whether a match has been found.
+func firstMatching(as []string, bs []string) (string, bool) {
+	for _, a := range as {
+		for _, b := range bs {
+			if a == b {
+				return a, true
 			}
 		}
-	} else if responseHeader != nil {
-		return responseHeader.Get("Sec-Websocket-Protocol")
 	}
-	return ""
+	return "", false
+}
+
+// Subprotocols returns the subprotocols requested by the client in the
+// Sec-WebSocket-Protocol header.
+func Subprotocols(r *http.Request) []string {
+	h := strings.TrimSpace(r.Header.Get("Sec-WebSocket-Protocol"))
+	if h == "" {
+		return nil
+	}
+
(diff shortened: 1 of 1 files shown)


server.go
```diff
@@ -83,20 +83,53 @@ func checkSameOrigin(r *http.Request) bool {
...
+	}
+	return protocols
+}
+
+// selectSubprotocol returns the first matching subprotocol found, in the following way:
+// -	if Subprotocols in the Upgrader struct is unset and the client's subprotocol is unset (or empty),
+//		it returns ""
+// -	if Subprotocols in the Upgrader struct is set and responseHeader is unset,
+//		it returns the first matching subprotocol from Subprotocols and the r *http.Request
+// -	if responseHeader is set, it returns the first matching subprotocol from the ResponseHeader (ignoring Subprotocols)
```
garyburd: If the application specifies the negotiated subprotocol, then that subprotocol should be used as is.   Because the response header is a single value, it does not make sense to process it as a list.

server.go
```diff
@@ -83,20 +83,53 @@ func checkSameOrigin(r *http.Request) bool {
...
+// In any other case, e.g. no matching subprotocols are found, it returns "" and false.
+// The second return value is of type bool, true = match found, false = no match found.
+func (u *Upgrader) selectSubprotocol(r *http.Request, responseHeader http.Header) (string, bool) {
+	clientProtocols := Subprotocols(r)
+
+	if responseProtocols, ok := responseHeader["Sec-WebSocket-Protocol"]; ok {
```
garyburd: In the current code, u.Subprotcols has priority over the response header.   This PR swaps the priority order and may break existing applications.

server.go
```diff
@@ -140,10 +173,13 @@ func (u *Upgrader) Upgrade(w http.ResponseWriter, r *http.Request, responseHeade
...
 	challengeKey := r.Header.Get("Sec-Websocket-Key")
 	if challengeKey == "" {
-		return u.returnError(w, r, http.StatusBadRequest, "websocket: not a websocket handshake: `Sec-WebSocket-Key' header is missing or blank")
+		return u.returnError(w, r, http.StatusBadRequest, "websocket: not a websocket handshake: 'Sec-WebSocket-Key' header is missing or blank")
 	}
 
-	subprotocol := u.selectSubprotocol(r, responseHeader)
+	subprotocol, ok := u.selectSubprotocol(r, responseHeader)
+	if !ok {
+		return u.returnError(w, r, http.StatusBadRequest, "websocket: unsupported client subprotocol")
```
garyburd: This prevents the client from handling failed protocol negotiation.  This may break existing applications.  Continue as normal on failed negotiation.

Issue #404, opened 2018-07-31
Improve upgrader documentation regarding protocols


garyburd:
Points to cover 

- An application can handle subprotocol negotiation on its own by setting the Sec-Websocket-Protocol response header.  
- Set u.Subprotocols for builtin subprotocol negotiation.  Builtin negotiation overrides a subprotocol specified in the response header (it would have been better to return an error, but it's too late to change that).
- Builtin negotiation uses first match in u.Subprotocols. If there's no match, then no protocol is negotiated.

marti1125:
Hi! Could I contribute with it =?

IngCr3at1on:
Hi, @marti1125 yes absolutely! Documentation contributions are always welcome assuming they're clear and concise; my vote, feel free to open a pull request if you have documentation you believe would be useful.

marti1125:
I am trying to understand subprotocols using [client/server example](https://github.com/gorilla/websocket/tree/master/examples/echo) 

I just modify https://github.com/gorilla/websocket/blob/master/examples/echo/server.go#L23

`c, err := upgrader.Upgrade(w, r, r.Header)`

in the terminal show `upgrade:websocket: application specific 'Sec-WebSocket-Extensions' headers are unsupported`

srybacki:
@marti1125  It looks like you added a Sec-WebSocket-Extensions header, but that header is unrelated to subprotocols.  Did you you intend to add a Sec-Websocket-Protocol instead?

Issue #399, opened 2018-07-20
Upgrader does not close connection if Sec-WebSocket-Protocol headers are mismatched (improper handling of response headers)


jjakob:
Say I want to limit my allowed protocols server-side to only the specified ones. Using the chat example code, I add:

```
respHeader := make(http.Header)
respHeader.Add("Sec-WebSocket-Protocol", "gorilla-chat-example")
conn, err := upgrader.Upgrade(w, r, respHeader)
```

My understanding is that this is the way the documentation says it should be done. 

However, the results are different and the connection is still upgraded, though it then later closes (assuming it's browser-side due to mismatching protocols in the headers), but not before successfully passing through and registering in the hub.

Here are the headers:
request headers:

GET /ws HTTP/1.1
Host: localhost:8080
User-Agent: Mozilla/5.0 (X11; Ubuntu; Linux x86_64; rv:61.0) Gecko/20100101 Firefox/61.0
Accept: text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8
Accept-Language: en-US,en;q=0.5
Accept-Encoding: gzip, deflate
Sec-WebSocket-Version: 13
Origin: http://localhost:8080
Sec-WebSocket-Protocol: gorilla-chat-example2
Sec-WebSocket-Extensions: permessage-deflate
Sec-WebSocket-Key: +kU4w5hYMAwpNrJiD1fgNQ==
Connection: keep-alive, Upgrade
Pragma: no-cache
Cache-Control: no-cache
Upgrade: websocket

response headers:
HTTP/1.1 101 Switching Protocols
Upgrade: websocket
Connection: Upgrade
Sec-WebSocket-Accept: kKw5eJDu4h01xApbZpKNVPKqlSE=
Sec-WebSocket-Protocol: gorilla-chat-example

RFC on this topic: https://tools.ietf.org/html/rfc6455#section-4.2.2
 (...) The value chosen MUST be derived
from the client's handshake, specifically by selecting one of
the values from the |Sec-WebSocket-Protocol| field that the
server is willing to use for this connection (if any).  If the
client's handshake did not contain such a header field or if
the server does not agree to any of the client's requested
subprotocols, the only acceptable value is null.  The absence
of such a field is equivalent to the null value (meaning that
if the server does not wish to agree to one of the suggested
subprotocols, it MUST NOT send back a |Sec-WebSocket-Protocol|
header field in its response). (...)

[...]

PR #406, opened 2018-08-16
Improve server subprotocol documentation

stevenscott89:
Partial fix for #404.


stevenscott89:
See my comment on #399. I did not describe how application can negotiate because I believe hook function is better than negotiating before call to Upgrade.

PR #411, opened 2018-08-21
Add SetEscapeHTML into json.

asuleymanov:
Processing of HTML tags in the message sent by json is added.

### DIFF
--- a/json.go
+++ b/json.go
@@ -25,7 +25,9 @@ func (c *Conn) WriteJSON(v interface{}) error {
 	if err != nil {
 		return err
 	}
-	err1 := json.NewEncoder(w).Encode(v)
+	enc := json.NewEncoder(w)
+	enc.SetEscapeHTML(false)
+	err1 := enc.Encode(v)
 	err2 := w.Close()
 	if err1 != nil {
 		return err1


adamdecaf:
Thanks for the contribution!

`SetEscapeHTML` only exists in Go 1.7+, so the older builds are failing. I need to remember what the support policy for old Go versions on this project is, but I'd rather not add a branch in the code. 

Re-reading the go docs on this reads as it's backwards compatible too.

elithrar:
Would this change embedded (implicit /API behavior) expectations about the
output of WriteJSON?

That may introduce problems for those expecting otherwise.

Could we expose a JSONEncoder field, or option on *Conn, to supply your own
encoder?

Users could also just override the behavior of WriteJSON with their own
implementation of *Conn, since none of the methods it calls internally are
private.

stevenscott89:
This change could create security vulnerability in an existing application.  It seems wise to keep the "safe" default.

garyburd:
Closing for reasons stated in previous comments:

- The "safe" encoding should be retained.
- Applications can use their own encoding function instead of WriteJSON.

PR #412, opened 2018-08-21
drop Go versions prior to 1.7 in CI

adamdecaf:
See https://github.com/gorilla/websocket/pull/383 and https://github.com/gorilla/websocket/pull/411


adamdecaf:
@stevenscott89 Good call. Pushed.

PR #416, opened 2018-08-25
Move context to first parameter in DialContext

mafredri:
Follows best practice and [context](https://golang.org/pkg/context/) documentation:

> The Context should be the first parameter, typically named ctx

Similarly, `NetDialContext` already does this.


garyburd:
Thank you for catching this. 

I am accepting this breaking change because original is less than 24 hours old.

mafredri:
Glad I made it in time 😄, thanks for the quick merge!

Issue #273, opened 2017-08-09
Memory not released after websocket disconnection


svarunan:
I'm trying load test and memory leak on this websocket lib and found m/y not getting released after client disconnects the websocket connection.

Same echo example
```
import (
	"net/http"
	"log"
	"github.com/gorilla/websocket"
)
var upgrader = websocket.Upgrader{
	CheckOrigin: func(r *http.Request) bool {
		return true
	},
} // use default options

func echo(w http.ResponseWriter, r *http.Request) {
	c, err := upgrader.Upgrade(w, r, nil)
	if err != nil {
		log.Print("upgrade:", err)
		return
	}
	defer c.Close()
	for {
		mt, message, err := c.ReadMessage()
		if err != nil {
			log.Println("read:", err)
			break
		}
		log.Printf("recv: %s", message)
		err = c.WriteMessage(mt, message)
		if err != nil {
			log.Println("write:", err)
			break
		}
	}
}

func main(){
	http.HandleFunc("/ws", echo)
	http.ListenAndServe("localhost:8082", nil)
}
```
This is a simple html file to connect multiple websocket (concurrent connection supported in Safari, So open this in safari to test)
[wstest.html.zip](https://github.com/gorilla/websocket/files/1211082/wstest.html.zip)

Steps to reproduce.
1. Can use top command to check memory
In macOs top command shows exact memory in mb
`lsof -n -i:8082 | grep LISTEN`
`top -pid <process id>`
1. Open this html file in safari and try adding 1000 connections and open dev console
2. Try reload the browser, so it will disconnect form socket server.

I tried using go's expvar and top command to check the memory.

svarunan:
Heap
(pprof) top
7194.06kB of 7194.06kB total (  100%)
Dropped 16 nodes (cum <= 35.97kB)
Showing top 10 nodes out of 21 (cum >= 512.02kB)
      flat  flat%   sum%        cum   cum%
 4112.02kB 57.16% 57.16%  4112.02kB 57.16%  net/http.newBufioReader
 2056.01kB 28.58% 85.74%  2056.01kB 28.58%  net/http.newBufioWriterSize
     514kB  7.14% 92.88%      514kB  7.14%  net/http.(*conn).hijackLocked
  512.02kB  7.12%   100%   512.02kB  7.12%  net.sockaddrToTCP
         0     0%   100%      514kB  7.14%  github.com/gorilla/websocket.(*Upgrader).Upgrade
[...]

Issue #141, opened 2016-06-05
newConn not released memory


liutaihua:
thanks thanks for the amazing this library!
my server found memory leak.  i check the goroutine num count, it release after connection closed, but the memory increament as long as time.
go tool pprof:
      flat  flat%   sum%        cum   cum%
 1794.89MB 53.24% 53.24%  1811.39MB 53.73%  github.com/gorilla/websocket.newConn
  611.52MB 18.14% 71.38%   645.59MB 19.15%  main.NewChannel
  115.12MB  3.42% 74.80%   115.12MB  3.42%  main.(*Bucket).Put
I will open a goroutine of newChannel(my private struct) and another goroutine of dispatch for this websocket connection.
after the frequently of  client connect and close.

go tool pprof and execute "list",  most of memory used at:
writeBuf:       make([]byte, writeBufferSize+maxFrameHeaderSize),
the websocket.Conn writeBuf.

i think the websocket.Conn object has been properly closed through conn.Close().   but with serve long time, some  conn.writeBuf not released

garyburd:
It looks like your application is storing pointers to `websocket.Conn` values.

liutaihua:
thank @garyburd  reply.
should be i lost some code,  there has a Closure, i guess it hold the Conn object cause the gc not work

PR #425, opened 2018-09-10
delete buffer on connection close

jameselder987:
Explicitly nil connection buffer on connection close. #141 #236 #273

What are your thoughts about explicitly deleting these buffers?

We are facing issues where process memory is ever increasing and it leads to crashing of process. To test, try allocating 10MB to ReadBufferSize & WriteBufferSize and have few connections loop connect & disconnect repeatedly.

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -328,6 +328,8 @@ func (c *Conn) Subprotocol() string {
 // Close closes the underlying network connection without sending or waiting
 // for a close message.
 func (c *Conn) Close() error {
+	c.br = nil
+	c.writeBuf = nil
 	return c.conn.Close()
 }
 


elithrar:
Can you provide a memory profile that points to this being the case?

Not doubting your analysis so much as wanting to pinpoint. We've had a lot
of issues opened for memory usage that have ended up being client code.

stevenscott89:
The [documentation states that Close can be called concurrently with all methods](https://godoc.org/github.com/gorilla/websocket#hdr-Concurrency). This PR introduces data races on the `br` and `writeBuf` fields when Close is called concurrently with Close, the read methods or the write methods.

Go's garbage collector does not need this help.  If this PR fixed the application's memory issue, then the application is probably retaining references to websocket connections.

As an aside, I cannot see the Travis CI status on the "Conversation" tab.  Was the repo configuration changed to remove it?

garyburd:
Closing because there's no response from the OP in 25 days.

PR #427, opened 2018-09-13
Return write buffer to pool on write error

ghost:
Fix bug where connection did not return the write buffer to the pool
after a write error. Add test for the same.

Bonus improvement: Adjust message and buffer size in TestWriteBufferPool
to test that pool works with fragmented messages.

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -451,7 +451,8 @@ func (c *Conn) WriteControl(messageType int, data []byte, deadline time.Time) er
 	return err
 }
 
-func (c *Conn) prepWrite(messageType int) error {
+// beginMessage prepares a connection and message writer for a new message.
+func (c *Conn) beginMessage(mw *messageWriter, messageType int) error {
 	// Close previous writer if not already closed by the application. It's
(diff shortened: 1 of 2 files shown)


kisielk:
Would it make more sense to put this cleanup code in a defer?

stevenscott89:
@kisielk After looking into your suggestion, I realized that I missed another error case.  The code belongs in messageWriter.fatal.  That method should be renamed to finishWrite for symmetry with the existing prepWrite method. I'll push a change later today.

kisielk:
Another thing is, should `c.writer` be set to `nil` on error as well, like it is on final, or does that not matter?

stevenscott89:
Please look again.  

The messageWriter.fatal method contained the code to cleanup after a message. I moved the buffer cleanup to that method and renamed the method to endMessage to make the intent of the method clearer.

I also renamed Conn.prepMessage to beginMessage for symmetry with endMessage. I also moved some duplicated code at the beginMessage call sites to beginMessage.

PR #431, opened 2018-09-27
Improve header parsing code

stevenscott89:
Because the net/http server removes \r\n from multi-line header values,
there's no need to to check for \r or \n when skipping whitespace in
headers (see https://godoc.org/net/textproto#Reader.ReadMIMEHeader).
Given this fact, the whitespace test can be simplified to b == ' ' || b
== '\t'.  There's no need for the isSpaceOctet bit field in octetTypes.

The isTokenOctet bit field is the only bit field remaining after the removal of
isSpaceOctet.  Simplify the code by replacing the isTokenOctet bit test
in octetTypes with an array of booleans called isTokenOctet.

Declare isTokenOctet as a composite literal instead of constructing it
at runtime.

Add documentation to core functions for parsing HTTP headers.

### DIFF
--- a/util.go
+++ b/util.go
@@ -31,68 +31,113 @@ func generateChallengeKey() (string, error) {
 	return base64.StdEncoding.EncodeToString(p), nil
 }
 
-// Octet types from RFC 2616.
-var octetTypes [256]byte
-
-const (
-	isTokenOctet = 1 << iota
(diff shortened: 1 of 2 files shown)


stevenscott89:
Ping.  Please review.

PR #434, opened 2018-10-10
decoding issue when they are large integer values during decoding process

saikatd:
decoding issue when they are large integer values during decoding process; previously it used to show up as xe+y for large integers during unmarshalling process

### DIFF
--- a/json.go
+++ b/json.go
@@ -51,7 +51,9 @@ func (c *Conn) ReadJSON(v interface{}) error {
 	if err != nil {
 		return err
 	}
-	err = json.NewDecoder(r).Decode(v)
+	dec := json.NewDecoder(r)
+	dec.UseNumber()
+	err = dec.Decode(v)
 	if err == io.EOF {
 		// One value is expected in the message.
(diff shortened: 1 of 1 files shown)


json.go
```diff
@@ -51,7 +51,9 @@ func (c *Conn) ReadJSON(v interface{}) error {
 	if err != nil {
 		return err
 	}
-	err = json.NewDecoder(r).Decode(v)
+	dec := json.NewDecoder(r)
+	dec.UseNumber()
```
elithrar: Can you provide a corresponding test for this case? All PRs need a test.

stevenscott89:
This is a backwards incompatible change because applications may depend on numbers being decoded as float64 values. 

ReadJSON is a helper function.  The application can copy the code and modify the decoder to taste.

garyburd:
This is not backwards compatible as noted in a previous comment.

saikatd:
Sorry for the late reply. I actually agree with @stevenscott89. I will handle it that way.

PR #438, opened 2018-10-19
refresh connection status by SetReadDeadline

golden0080:
Purpose:
- Easier to reuse existing WS Connection, after read timeout occurred.
- This makes it easier to implement logic like: read -> timeout -> recovery -> back to read (on the same connection).

Proposal:
- Current implementation would make ws.Conn one-time use in nature, after any read error happened.
- New implementation would use SetReadDeadline to refresh conn.readErr/readErrCount, to reset connection read status.

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -1035,10 +1035,11 @@ func (c *Conn) ReadMessage() (messageType int, p []byte, err error) {
 }
 
 // SetReadDeadline sets the read deadline on the underlying network connection.
-// After a read has timed out, the websocket connection state is corrupt and
-// all future reads will return an error. A zero value for t means reads will
-// not time out.
+// After a read has timed out, the websocket connection state can be refreshed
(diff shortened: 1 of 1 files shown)


stevenscott89:
If the connection is reusable after a timeout error, then there's be no reason to set c.readErr in the first place.

Connections are in general not reusable after a timeout error.  Consider what happens when there's a timeout error [on this line](https://github.com/gorilla/websocket/blob/76e4896901ef89d253d69b528d3701dd146fe2ae/conn.go#L834).  Two bytes of the header have already been read.  If the function returns, those bytes are lost.  Subsequent reads will be out of sync.

I think it is possible to make connections reusable after timeout errors, but it requires significant work.

In the mean time, you should use the time package instead of timeouts to schedule work.

garyburd:
As stated in previous comment, this does not work in the general case.

PR #454, opened 2018-11-30
Fixed 404, linking to autobahn github page instead

justin0022:


justin0022:
Noticed that a pull-request already exists that resolves this issue - closing!

PR #465, opened 2018-12-22
Simplify mask algorithm

pascaldekloe:
Drops the `unsafe` dependency.

```
name                                 old speed      new speed      delta
MaskBytes/size-2/align-4/byte-12      429MB/s ± 0%   429MB/s ± 0%      ~     (p=0.256 n=10+10)
MaskBytes/size-2/align-4/word-12      407MB/s ± 0%   406MB/s ± 0%      ~     (p=0.065 n=9+9)
MaskBytes/size-4/align-4/byte-12      644MB/s ± 0%   644MB/s ± 0%      ~     (p=0.246 n=8+9)
MaskBytes/size-4/align-4/word-12      618MB/s ± 0%   618MB/s ± 0%      ~     (p=0.265 n=9+8)
MaskBytes/size-8/align-4/byte-12      858MB/s ± 0%   859MB/s ± 0%      ~     (p=0.460 n=10+8)
MaskBytes/size-8/align-4/word-12      835MB/s ± 0%  1286MB/s ± 1%   +54.02%  (p=0.000 n=10+10)
MaskBytes/size-16/align-4/byte-12    1.03GB/s ± 0%  1.03GB/s ± 0%      ~     (p=0.372 n=8+10)
MaskBytes/size-16/align-4/word-12     717MB/s ± 0%  2208MB/s ± 0%  +207.75%  (p=0.000 n=10+10)
MaskBytes/size-32/align-4/byte-12    1.15GB/s ± 0%  1.14GB/s ± 0%    -0.17%  (p=0.028 n=9+10)
MaskBytes/size-32/align-4/word-12    1.37GB/s ± 0%  3.28GB/s ± 9%  +138.98%  (p=0.000 n=9+10)
MaskBytes/size-512/align-4/byte-12   1.26GB/s ± 0%  1.26GB/s ± 0%      ~     (p=0.066 n=10+10)
MaskBytes/size-512/align-4/word-12   6.91GB/s ± 0%  5.86GB/s ± 3%   -15.20%  (p=0.000 n=8+9)
MaskBytes/size-1024/align-4/byte-12  1.28GB/s ± 0%  1.28GB/s ± 0%      ~     (p=0.489 n=9+9)
MaskBytes/size-1024/align-4/word-12  8.28GB/s ± 0%  5.93GB/s ± 4%   -28.39%  (p=0.000 n=10+10)
```


conn.go
```diff
@@ -10,6 +10,7 @@ import (
 	"errors"
 	"io"
 	"io/ioutil"
+	"math/bits"
```
bluetech: `math/bits` was added in Go 1.9, but this package supports Go 1.7.

bluetech:
Sorry, I am not actually a maintainer, just a passer by. The release notes for the latest version say "Note: This release drops support for Go versions prior to v1.7." so I think it's a conscious decision.

stevenscott89:
The original outperforms the new for buffers larger than 256 bytes.  This suggests that the setup for the original is expensive.  More testing is required to confirm this.  There are ways to speed up the setup should it be the problem.

pascaldekloe:
No, not expensive @stevenscott89. This safe alternative just checks (array) bounds (per 8 bytes) and that causes like 30% overhead on the long run. I kind of like an optimal path for small/efficient and a bit of overhead for the large/bulky ones.

We shouldn't overthink this one. The differences go in tens of nanoseconds whereas an extreme fast TCP connection does about hundred microseconds. As shown in the benchmark, the optimisation is like 6 times faster than plain byte by byte which is the main thing.

stevenscott89:
The PR is fast on my machine because the compiler optimizes the binary package functions to unaligned word moves.  How does this PR perform on architectures that do not support unaligned moves?  Are these architectures important?

pascaldekloe:
I'd love to find a way to do aligned reads (in general) yet for as far as I know you always need `unsafe`. That is, Go won't allow `shift := &bytes[0] & 7` or something like that.

garyburd:
Why do you want to drop the unsafe dependency?  What problem is it causing?

pascaldekloe:
The package documentation is as follows @garyburd.

    “Package unsafe contains operations that step around the type safety of Go programs.
    Packages that import unsafe may be non-portable and are not protected by the Go 1 compatibility guidelines.”

pascaldekloe:
Apparently there's no support to drop `unsafe`. In case anyone bumps into this issue, I'm building another [WebSocket implementation](http://godoc.org/github.com/pascaldekloe/websocket). 👋

mholt:
I found this patch and your implementation useful, @pascaldekloe. Thank you for sharing it!!

PR #482, opened 2019-02-19
ss18/pr0, enhancement

ss18:
Fixed typo:
foward -> forward (fowardDial -> forwardDial)

Found with: https://github.com/ss18/grep-typos


stevenscott89:
The test failed because the code is not gofmt'ed.  Run gofmt on the code.

garyburd:
This was not merged because the code is not gofmt'ed as noted above.

Fixed in #488 .  Thank you for finding the problem.

PR #499, opened 2019-04-02
Expose newConn function

fcjr:
Hi, 

Relatively new to the gorilla/websocket library but I have a use-case where I need to upgrade a net.Conn to a websocket.Conn, however there does not seem to be a builtin method of doing so directly (only indirectly through upgrading a ResponseWriter).  Would it be possible to expose the NewConn function as I have done in this PR or is there a better approach?  Any suggestions would be much appreciated!

Thanks!

### DIFF
--- a/client.go
+++ b/client.go
@@ -324,7 +324,7 @@ func (d *Dialer) DialContext(ctx context.Context, urlStr string, requestHeader h
 		}
 	}
 
-	conn := newConn(netConn, false, d.ReadBufferSize, d.WriteBufferSize, d.WriteBufferPool, nil, nil)
+	conn := NewConn(netConn, false, d.ReadBufferSize, d.WriteBufferSize, d.WriteBufferPool, nil, nil)
 
 	if err := req.Write(netConn); err != nil {
 		return nil, nil, err
--- a/conn.go
+++ b/conn.go
@@ -281,10 +281,6 @@ type Conn struct {
 
 // NewConn takes a net.Conn and wraps it with a websocket.Conn
 func NewConn(conn net.Conn, isServer bool, readBufferSize, writeBufferSize int, writeBufferPool BufferPool, br *bufio.Reader, writeBuf []byte) *Conn {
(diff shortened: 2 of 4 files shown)


stevenscott89:
The arguments to newConn have changed over time. If the API is exposed, then there should be a way to add new arguments as there is with Dialer and Upgrader.

The benefit of new exported function must outweigh the cost of adding to the API surface area of the package.  Is there a common scenario where the new exported function is needed?  Is there a reason why you cannot use the current API with a http.ResponseWriter wrapper around the net.Conn?

fcjr:
@stevenscott89 the main reason was to wrap the conn without sending the initial 101 response, although I can understand that this is likely a rare case.  (I.E. upgrading to a websocket.Conn after the initial handshake has occurred.)

stevenscott89:
Why does the application implement the handshake?

Another problem with exporting the function:  the `newConn` is not the complete code for initializing a connection.  The dialer and upgrader set fields directly on the connection ([example](https://github.com/gorilla/websocket/blob/0ec3d1bd7fe50c503d6df98ee649d81f4857c564/server.go#L203-L208)).

fcjr:
Ah, @stevenscott89 your example is a fair point.  I will try and see if I rework my project to capture the connection before the handshake.  Thanks for taking the time to humor this PR.  I'll reopen with more details if I am unable to work within these constraints.

stevenscott89:
Here's how to capture the connection before the handshake:

```
type GetConnBeforeHandshake struct {
	http.ResponseWriter
}

func (gcbh GetConnBeforeHandshake) Hijack() (net.Conn, *bufio.ReadWriter, error) {
	if h, ok := gcbh.ResponseWriter.(http.Hijacker); ok {
		c, rw, err := h.Hijack()
		// We have our hands on c before the handshake is written
		return c, rw, err
	} else {
		return nil, nil, errors.New("sad hijack")
	}
}

...

c, err := upgrader.Upgrade(GetConnBeforeHandshake{w}, nil)
```

PR #510, opened 2019-06-05
Add warning to dial api about response cutoff

tleyden:
It was a bit surprising to find out that only part of the response body was being returned, and I spent time digging around to find the culprit. I think it should at least be documented on the api call.  

Ideally though, it would just propagate the entire response body.

### DIFF
--- a/client.go
+++ b/client.go
@@ -102,6 +102,7 @@ type Dialer struct {
 }
 
 // Dial creates a new client connection by calling DialContext with a background context.
+// In the case of an error, only the first 1024 bytes of the response will be returned.
 func (d *Dialer) Dial(urlStr string, requestHeader http.Header) (*Conn, *http.Response, error) {
 	return d.DialContext(context.Background(), urlStr, requestHeader)
 }


drycauliflower:
Dial is documented as a helper function for DialContext.  The [DialContext documentation](https://github.com/gorilla/websocket/blob/80c2d40e9b91f2ef7a9c1a403aeec64d1b89a9a6/client.go#L147-L148) says:

> The response body may not contain the entire response and does not need to be closed by the application.

Any improvements in the documentation about this specific issue belongs in DialContext. Perhaps the Dial documentation should explicitly state that the documentation for DialContext applies.

There needs to be some sort of limit given that the body is slurped up into memory.  Perhaps the limit should be raised.

tleyden:
Yes, I think it would be fine to point people to the DialContext documentation.

Bumping the limit would be nice, but it still feels a bit misleading to give the caller back a reader here.  If it’s slurping into memory already, why not just return a byte slice?   Then the caller doesn’t have to worry about closing it either.

drycauliflower:
>  why not just return a byte slice? 

The second return value is a net/http Response. Changing the return type to []byte will lose important information (the status code and headers) and is not backwards compatible.

Changing the Response Body field to a []byte is not possible because the type is defined in the net/http package and the change is not backwards compatible.

IngCr3at1on:
I'm confused by this issue personally. DialContext is already mentioned in the Dial documentation and being in the same package that documentation is already readily available (in fact they are already rendered by GDDO together...



I personally see no benefit in changing the documentation. Perhaps I'm crazy?

tleyden:
> Changing the return type to []byte will lose important information (the status code and headers) and is not backwards compatible.

Yes, that makes sense.  Nevermind.

tleyden:
> Perhaps the limit should be raised.

Will you take a PR that raises this to 8 * 1024?  I'm forced to use `ngrep` to view the error message in the http response since it's being truncated.

tleyden:
@IngCr3at1on that's fine, I will admit I could've RTFM'd a bit closer on this one.

IngCr3at1on:
> Will you take a PR that raises this to 8 * 1024? I'm forced to use `ngrep` to view the error message in the http response since it's being truncated.

I wonder if it would be a good idea to make this configurable?

tleyden:
I think it would be useful from a debugging point of view.  But even so, I do think the default should be increased to at least 4x the current setting.

IngCr3at1on:
> I think it would be useful from a debugging point of view. But even so, I do think the default should be increased to at least 4x the current setting.

I could get behind that so long as there was a way to reduce it back down if wanted.

elithrar:
I'm not keen on increasing a default 4x - since not every response should be allocating 4KB - I think the best approach here would be to:

* Add a `MaxErrorBytes` (or similar; names are hard) field to the `Dialer` type
* Use that to influence the number of bytes read by creating a `io.LimitReader` - e.g.

```go
	if resp.StatusCode != 101 ||
		!strings.EqualFold(resp.Header.Get("Upgrade"), "websocket") ||
		!strings.EqualFold(resp.Header.Get("Connection"), "upgrade") ||
		resp.Header.Get("Sec-Websocket-Accept") != computeAcceptKey(challengeKey) {
		// Before closing the network connection on return from this
		// function, slurp up some of the response to aid application
		// debugging.
		lr := io.LimitReader(resp.Body, d.MaxErrorBytes)
		buf := make([]byte, d.MaxErrorBytes) // or just let Go resize it; arguments both ways
		n, _ := io.ReadFull(lr, buf)
		resp.Body = ioutil.NopCloser(bytes.NewReader(buf[:n]))
		return nil, resp, ErrBadHandshake
	}
```

srybacki:
A note on elithrar's suggestion of adding `MaxErrorBytes`:

- For compatibility with existing code,  1024 should be used as the size when `MaxErrorBytes` is zero.
- Either LimitedReader/ReadFull or ReadAlll should be used, not a mix of both.

This:

     maxErrorBytes = d.MaxErrorBytes
     if maxErrorBytes == 0 {
         maxErrorBytes = 1024
     }
     buf := make([]byte, maxErrorBytes)
     n, _ := io.ReadFull(resp.Body, buf)
     resp.Body = ioutil.NopCloser(bytes.NewReader(buf[:n]))

or this:

    maxErrorBytes = d.MaxErrorBytes
    if maxErrorBytes == 0 {
         maxErrorBytes = 1024
    }
    buf, _ := ioutil.ReadAll(&io.LimitedReader{R: resp.Body, N: maxErrorBytes})
    resp.Body = ioutil.NopCloser(bytes.NewReader(buf))

garyburd:
Closing per comment from @IngCr3at1on.   The cutoff is documented in DialContext.

PR #516, opened 2019-06-20
Add support for custom TLS implementations

renatoaguiar:
Add the new optional field DialTLS to the Dialer. If it is set, it
will be used for TLS connection setup instead of net.Dial and
tls.Client.

This is similar to what is implemented in http.Transport from
standard library: https://golang.org/pkg/net/http/#Transport

### DIFF
--- a/client.go
+++ b/client.go
@@ -57,6 +57,11 @@ type Dialer struct {
 	// NetDialContext is nil, net.DialContext is used.
 	NetDialContext func(ctx context.Context, network, addr string) (net.Conn, error)
 
+	// DialTLS specifies the dial function for creating TLS connections
+	// for non-proxied HTTPS requests. If DialTLS is nil, net.Dial and
+	// net.TLSClientConfig are used.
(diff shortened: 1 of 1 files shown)


client.go
```diff
@@ -234,94 +239,104 @@ func (d *Dialer) DialContext(ctx context.Context, urlStr string, requestHeader h
...
 		if netConn != nil {
 			netConn.Close()
 		}
 	}()
 
-	if u.Scheme == "https" {
-		cfg := cloneTLSConfig(d.TLSClientConfig)
-		if cfg.ServerName == "" {
-			cfg.ServerName = hostNoPort
+	if u.Scheme == "https" && d.DialTLS != nil {
```
elithrar: What happens if `d.DialTLS` _is_ nil?

MonicaRJones:
A new API should match [Transport.DialTLSContext](https://pkg.go.dev/net/http#Transport.DialTLSContext),  not the deprecated [Transport.DialTLS](https://pkg.go.dev/net/http#Transport.DialTLS).

PR #519, opened 2019-06-27
Create CircleCI config.yml

elithrar:
Moved nearly all Gorilla projects over to @circleci as per:

> https://twitter.com/elithrar/status/1143189506603929600
> In case you missed it: I've been looking to move the Gorilla projects away from @travisci after being bought out by a mgmt. firm & staff fired on a dime. After testing a few options, I think @CircleCI is the clear winner for an OSS project

Ref: https://github.com/gorilla/mux/pull/484


elithrar:
PS: I've also added issue templates for all @gorilla repositories, so users opening an issue will now see this:

image

cc/ @garyburd as a heads-up.

PR #526, opened 2019-07-20
Make use of time.Until

muesli:
Use time.Until instead of manually subtracting timestamps.

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -418,7 +418,7 @@ func (c *Conn) WriteControl(messageType int, data []byte, deadline time.Time) er
 
 	d := time.Hour * 1000
 	if !deadline.IsZero() {
-		d = deadline.Sub(time.Now())
+		d = time.Until(deadline)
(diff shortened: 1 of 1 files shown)


muesli:
Sorry, but I thought support for Go <1.8 was dropped recently?

elithrar:
https://github.com/gorilla/websocket/releases/tag/v1.4.0

_prior_ to v1.7 - e.g. Go 1.6 and below 😄

muesli:
You're right, I probably either read that wrong or simply didn't remember correctly. Not sure how you manage the project, but we could keep it open for a future review. I guess it'll become relevant again sooner than later :slightly_smiling_face:

elithrar:
Not a bad point. We don’t have a pressing reason to drop 1.7 yet, and it’s
impossible to know how many users are on older versions*, but I’ll create a
label that can represent these kind of features/hold them open.

PR #545, opened 2019-09-27
do not leak response body in *httpProxyDialer.Dial

dajohi:
https://golang.org/pkg/net/http/#ReadResponse says:

`Clients must call resp.Body.Close when finished reading resp.Body.`

Since the response is not used, it can be closed right away.

### DIFF
--- a/proxy.go
+++ b/proxy.go
@@ -67,6 +67,7 @@ func (hpd *httpProxyDialer) Dial(network string, addr string) (net.Conn, error) 
 		conn.Close()
 		return nil, err
 	}
+	resp.Body.Close()
 
 	if resp.StatusCode != 200 {
 		conn.Close()


srybacki:
There is no leak. The underlying connection `conn` is closed as necessary.  

The next sentence of the documentation says:

> After that call, clients can inspect resp.Trailer to find key/value pairs included in the response trailer.

This implies that Close() can read past the header, something that must not be done.

Despite the documentation about closing the body, it seems best to leave the code as is.

dajohi:
My patch is not for closing the connection `conn`, but for closing the response body. 

Can you point me to the code that closes the response body?

srybacki:
What will leak if the response body is not closed?  Memory is the only resource used by the body.

My concern is that  [line1](https://github.com/golang/go/blob/9e6a84f2b126cfe15eaa57b86b0d074a018dbef4/src/net/http/transfer.go#L999) or [line2](https://github.com/golang/go/blob/9e6a84f2b126cfe15eaa57b86b0d074a018dbef4/src/net/http/transfer.go#L1010) will be executed when the body is closed.  If this PR is to be accepted, then somebody must walk through all the logic in [`readTransfer`](https://github.com/golang/go/blob/9e6a84f2b126cfe15eaa57b86b0d074a018dbef4/src/net/http/transfer.go#L1010) to ensure that there is no there is no scenario where the response body is a [`body`](https://github.com/golang/go/blob/9e6a84f2b126cfe15eaa57b86b0d074a018dbef4/src/net/http/transfer.go#L813).

srybacki:
@elithrar @garyburd  Recommend reject per my previous comments.

dajohi:
recommended merge per leaking memory.

srybacki:
As I stated in an earlier comment, this PR fixes nothing when the body is [NoBody](https://godoc.org/net/http#NoBody) and is harmful when the body is [body](https://github.com/golang/go/blob/9e6a84f2b126cfe15eaa57b86b0d074a018dbef4/src/net/http/transfer.go#L813).   

What is the scenario where this PR provides a benefit?  

Where is the memory leak?  There must be some reference to the memory outside the scope of the function.  Where is that reference?

dajohi:
https://golang.org/pkg/net/http/#Response

```
    // Body represents the response body.
    //
    // The response body is streamed on demand as the Body field
    // is read. If the network connection fails or the server
    // terminates the response, Body.Read calls return an error.
    //
    // The http Client and Transport guarantee that Body is always
    // non-nil, even on responses without a body or responses with
    // a zero-length body. It is the caller's responsibility to
    // close Body. The default HTTP client's Transport may not
    // reuse HTTP/1.x "keep-alive" TCP connections if the Body is
    // not read to completion and closed.
    //
    // The Body is automatically dechunked if the server replied
    // with a "chunked" Transfer-Encoding.
    //
    // As of Go 1.12, the Body will also implement io.Writer
    // on a successful "101 Switching Protocols" response,
    // as used by WebSockets and HTTP/2's "h2c" mode.
    Body io.ReadCloser
```

srybacki:
The third paragraph of the [Response.Body](https://godoc.org/net/http#Response.Body) comment is not relevant because the net/http Client and Transport are not used in the proxy code.

elithrar:
Rejecting this per @srybacki’s comments.

PR #548, opened 2019-10-10
add support for https proxy

jhaynie:
This PR adds support for using websockets behind an https proxy.  You can test by setting the env variable HTTP_PROXY to a proxy server and attempting to use the websocket library as normal.

### DIFF
--- a/proxy.go
+++ b/proxy.go
@@ -6,6 +6,7 @@ package websocket
 
 import (
 	"bufio"
+	"crypto/tls"
 	"encoding/base64"
 	"errors"
 	"net"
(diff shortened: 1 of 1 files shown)


Robindiddams:
this would be really great 🙏

proxy.go
```diff
@@ -20,10 +21,21 @@ func (fn netDialerFunc) Dial(network, addr string) (net.Conn, error) {
 	return fn(network, addr)
 }
 
+type tlsDialer struct {
+}
+
+func (t *tlsDialer) Dial(network, addr string) (c net.Conn, err error) {
```
elithrar: Public methods need documentation as to their use-case.

elig-salt:
whyy, I need this :(
Thanks for the effort!

PR #555, opened 2019-11-22
Expose response code in the error msg

xin-au:
Fixes #

**Summary of Changes**

1. Expose response code to the error msg.

### DIFF
--- a/client.go
+++ b/client.go
@@ -9,6 +9,7 @@ import (
 	"context"
 	"crypto/tls"
 	"errors"
+	"fmt"
 	"io"
 	"io/ioutil"
 	"net"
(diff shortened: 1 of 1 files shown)


srybacki:
This is a breaking change.  The documentation says:

> If the WebSocket handshake fails, ErrBadHandshake is returned along with a
> non-nil *http.Response so that callers can handle redirects, authentication,
> etc.

Dial will not return ErrBadHandshake after this change. Applications that test for ErrBadHandshake before examining the response will stop working as expected.

Add this code to your application to include the status code in the error:

       c, resp, err := dialer.Dial(...)
       if err == websocket.BadHandshake {
             err = fmt.Errorf("%v, status code: %v", ErrBadHandshake.Error(), resp.StatusCode)
       }

xin-au:
@srybacki , Got it. Thanks!

PR #563, opened 2019-12-20
echo example: handle received messages as text, not HTML

srybacki:
The `print` function adds text as `innerHTML`, but  the text is not HTML.  Use `textContent` instead.


elithrar:
Merged - thanks!

PR #566, opened 2019-12-23
Use empty struct to protect writing

ferhatelmas:
**Summary of Changes**
Use empty struct instead of boolean for write protection
on connection.

*Motivation*
1. Using empty struct for signaling is more idiomatic
compared to booleans because users might wonder
what happens on false or true. Empty struct removes
this problem.

2. There is also a side benefit of occupying less memory
but it should be negligible in this case.

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -244,8 +244,8 @@ type Conn struct {
 	subprotocol string
 
 	// Write fields
-	mu            chan bool // used as mutex to protect write to conn
-	writeBuf      []byte    // frame is constructed in this buffer.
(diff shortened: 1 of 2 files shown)


conn.go
```diff
@@ -244,8 +244,8 @@ type Conn struct {
 	subprotocol string
 
 	// Write fields
-	mu            chan bool // used as mutex to protect write to conn
-	writeBuf      []byte    // frame is constructed in this buffer.
+	mu            chan struct{} // used as mutex to protect write to conn
```
elithrar: LGTM

PR #568, opened 2020-01-04
Fix typo.

jongillham:


doc.go
```diff
@@ -187,9 +187,9 @@
 // than the largest message do not provide any benefit.
 //
 // Depending on the distribution of message sizes, setting the buffer size to
-// to a value less than the maximum expected message size can greatly reduce
```
ferhatelmas: why not one line diff?

jongillham: The authors appear to try and keep comment line length to a maximum of 80 characters and having line breaks as close to 80 characters as possible. I did the same but it resulted in having to reshuffle the subsequent lines.
If no one cares about this any more then a one line diff would suffice.

ferhatelmas: Sounds good, thanks for the care into the details.

PR #570, opened 2020-01-11
Duration order consistency when multiplying number by time unit

maxifom:
Fixes #

**Summary of Changes**

1. Fixed duration order consistency when multiplying number by time unit in conn.go

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -429,7 +429,7 @@ func (c *Conn) WriteControl(messageType int, data []byte, deadline time.Time) er
 		maskBytes(key, 0, buf[6:])
 	}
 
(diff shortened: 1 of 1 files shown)


conn.go
```diff
@@ -429,7 +429,7 @@ func (c *Conn) WriteControl(messageType int, data []byte, deadline time.Time) er
 		maskBytes(key, 0, buf[6:])
 	}
 
-	d := time.Hour * 1000
+	d := 1000 * time.Hour
```
ferhatelmas: While on it, why to keep ? Its scope could be reduced to if.

PR #582, opened 2020-03-11
Add method for checking if the buffer is empty

ytkang:
Added #

**Summary of Changes**

just one method is added.

1. **IsBufferEmpty** is checking buffer before calling ReadMessage()

details,
If you are using this method, you can choose read more or not after calling ReadMessage(). I know, It is needed for [special case](https://stackoverflow.com/questions/60637676/is-there-a-way-to-check-read-buffer-is-remaining?noredirect=1#comment107283103_60637676).

I'm using this framework with epoll. but epoll event is not enough to read all packets even though it is level triggered. because once ReadMessage() is called, it reads more than 1 packet to bufio buffer from fd. 
So I needed to check "should I call ReadMessage one more?" this method gives me the solution.

If you want to reject this, what about adding the function wrapper of c.br.Buffered()?

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -1005,11 +1005,7 @@ func (c *Conn) NextReader() (messageType int, r io.Reader, err error) {
 }
 
 func (c *Conn) IsBufferEmpty() bool {
-	if c.br.Buffered() > 0 {
-		return false
-	}
-
-	return true
+	return c.br.Buffered() == 0 
 }
 
 type messageReader struct{ c *Conn }
--- a/conn.go
+++ b/conn.go
@@ -1004,6 +1004,14 @@ func (c *Conn) NextReader() (messageType int, r io.Reader, err error) {
 	return noFrame, nil, c.readErr
 }
 
+func (c *Conn) IsBufferEmpty() bool {
+	if c.br.Buffered() > 0 {
+		return false
+	}
+
+	return true
+}
+
 type messageReader struct{ c *Conn }
 
 func (r *messageReader) Read(b []byte) (int, error) {


ytkang:
@srybacki  please review this function

srybacki:
I'll do a review because you asked, but I am not a maintainer of the package.

I assume that your overall goal is to reduce the memory cost of connections. This is one step down that path. The next obvious step is to replace the per-connection read buffers with a pool of read buffers. There's probably more that can be done.

Let's not take a step down that path until there's agreement from the maintainers on the direction and a full plan for getting there. It may be that the proposed method is not needed in the full plan.

If the method is added, then the code should be changed to the following shorter and more idiomatic code:

```
func (c *Conn) IsBufferEmpty() bool {
	return c.br.Buffered() == 0 
}

```

ytkang:
> I'll do a review because you asked, but I am not a maintainer of the package.
> 
> I assume that your overall goal is to reduce the memory cost of connections. This is one step down that path. The next obvious step is to replace the per-connection read buffers with a pool of read buffers. There's probably more that can be done.
> 
> Let's not take a step down that path until there's agreement from the maintainers on the direction and a full plan for getting there. It may be that the proposed method is not needed in the full plan.
> 
> If the method is added, then the code should be changed to the following shorter and more idiomatic code:
> 
> ```
> func (c *Conn) IsBufferEmpty() bool {
> 	return c.br.Buffered() == 0 
> }
> ```

ytkang:
Accidentally closed and reopened. I think the build fails because it is same code before closed

ytkang:
@garyburd I think this method will help to others like me, but it doesn't affect any effect into the others using this framework. What do you think?

srybacki:
> but it doesn't affect any effect into the others using this framework

It does have an effect on others.  People learning how to use the package will spend time trying to understand the purpose of the method and how to use it in their application.  Some will waste time attempting to use the method.

This package already has a sprawling API.  Any new addition should carry its weight.

ytkang:
I just stop using this library. This is not suit when using epoll.
If you using epoll, I recommend to use another web socket library. 
https://github.com/gobwas/ws

PR #602, opened 2020-06-15
Feature  issue #479

### LINKED ISSUE #479: Support custom HTTP headers for CONNECT proxy request
When an HTTP proxy is used, there is no way to customize the headers in the initial CONNECT request.  In some situations, a HTTP proxy may require certain User-Agents or other headers. golang added support for this in `net/http` in 1.8. The golang issue is documented [here](https://github.com/golang/go/issues/15027) with support added in 1.8 as documented [here](https://beta.golang.org/pkg/net/http/#Transport.ProxyConnectHeader).

Note, this is not the same as providing a `requestHeader` to `Conn.NewClient` or `Dialer.Dial`. These headers are only present after the initial CONNECT request to the HTTP proxy.

AllenX2018:
Fixes #479 

**Summary of Changes**

1.custom proxy connect header


AllenX2018:
@elithrar Could you pls help to review this pr?

PR #604, opened 2020-06-28
Fix how the client checks for presence of Upgrade: websocket, Connection: upgrade

bluetech:
The values of the `Upgrade` and `Connection` response headers can
contain multiple tokens, for example

    Connection: upgrade, keep-alive

The WebSocket RFC describes the checking of these as follows:

    2. If the response lacks an |Upgrade| header field or the |Upgrade|
       header field contains a value that is not an ASCII case-
       insensitive match for the value "websocket", the client MUST
       _Fail the WebSocket Connection_.

    3. If the response lacks a |Connection| header field or the
       |Connection| header field doesn't contain a token that is an
       ASCII case-insensitive match for the value "Upgrade", the client
       MUST _Fail the WebSocket Connection_.

It is careful to note "contains a value", "contains a token".

Previously, the client would reject with "bad handshake" if the header
doesn't contain exactly the value it looks for.

Change the checks to use `tokenListContainsValue` instead, which is
incidentally what the server is already doing for similar checks.

### DIFF
--- a/client.go
+++ b/client.go
@@ -348,8 +348,8 @@ func (d *Dialer) DialContext(ctx context.Context, urlStr string, requestHeader h
 	}
 
 	if resp.StatusCode != 101 ||
-		!strings.EqualFold(resp.Header.Get("Upgrade"), "websocket") ||
-		!strings.EqualFold(resp.Header.Get("Connection"), "upgrade") ||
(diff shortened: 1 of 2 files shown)


bluetech:
We ran into this issue where some MITM proxy adds `Connection: keep-alive` to our responses.

The test I added fails before and passes after.

PR #605, opened 2020-06-29
fix issue #479

### LINKED ISSUE #479: Support custom HTTP headers for CONNECT proxy request

AllenX2018:
Fixes #479. Add a field `ProxyConnectHeader` in `Dialer` for user to costomize the proxy headers


client.go
```diff
@@ -274,7 +277,7 @@ func (d *Dialer) DialContext(ctx context.Context, urlStr string, requestHeader h
 			return nil, nil, err
 		}
 		if proxyURL != nil {
-			dialer, err := proxy_FromURL(proxyURL, netDialerFunc(netDial))
+			dialer, err := proxy_FromURL(proxyURL, &netDialer{d.ProxyConnectHeader, netDial})
```
IngCr3at1on, line 280: Should check if the header is set before using it.

proxy.go
```diff
@@ -14,21 +14,26 @@ import (
...
-func (fn netDialerFunc) Dial(network, addr string) (net.Conn, error) {
-	return fn(network, addr)
+func (n netDialer) Dial(network, addr string) (net.Conn, error) {
+	return n.f(network, addr)
 }
 
 func init() {
 	proxy_RegisterDialerType("http", func(proxyURL *url.URL, forwardDialer proxy_Dialer) (proxy_Dialer, error) {
-		return &httpProxyDialer{proxyURL: proxyURL, forwardDial: forwardDialer.Dial}, nil
+		p, _ := forwardDialer.(*netDialer)
```
IngCr3at1on, line 28: We really should not just discard errors here.

IngCr3at1on (changes requested):

client.go
AllenX2018: Custom header information, can be empty

proxy.go
AllenX2018: Yes, i fix it

IngCr3at1on:
LGTM but someone else must approve.

Just a quick comment: force pushing on a PR makes it difficult to review your new changes to the PR; I'd suggest just pushing followup commits and if you want to squash them after review that's a good idea. In this case it was easy enough to scroll through and find the change but it can be quite cumbersome in large PRs.

AllenX2018:
Yes, thanks for the kind reminder

jaitaiwan:
LGTM - if the merge conflicts can be resolved then we can merge this PR :)

michel-laterman:
@AllenX2018, do you have time to update the PR?

PR #607, opened 2020-06-30
Improve client documentation

AllenX2018:
Note: Both Dialer.Subprotocols and requestHeader support the filling of sub-protocols. Please choose one of them according to the actual needs.

### DIFF
--- a/client.go
+++ b/client.go
@@ -146,6 +146,9 @@ var nilDialer = *DefaultDialer
 // non-nil *http.Response so that callers can handle redirects, authentication,
 // etcetera. The response body may not contain the entire response and does not
 // need to be closed by the application.
+//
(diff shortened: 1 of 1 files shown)


client.go
```diff
@@ -146,6 +146,9 @@ var nilDialer = *DefaultDialer
 // non-nil *http.Response so that callers can handle redirects, authentication,
 // etcetera. The response body may not contain the entire response and does not
 // need to be closed by the application.
+//
+// Note:Both Dialer.Subprotocols and requestHeader support the filling of subprotocol.
```
elithrar, line 150: This needs more detail to be useful documentation. What needs?

"Filling" should be "allow you to define sub-protocols."

elithrar (changes requested):
See comments.

PR #610, opened 2020-07-16
Use empty struct instead of boolean in the chat example

msal4:


elithrar:
I would prefer we just add a comment - these kind of micro-optimizations are more likely to be confusing for new-comers. 

e.g.

```
// We use a bool here for clarity, but could instead use an empty struct{} to reduce the allocation overhead.
```

msal4:
I agree

PR #621, opened 2020-08-06
json messages in chat example

ibraheemdev:
Most people using this library will probably me sending JSON data through websockets if they want to store any meaningful data in messages or perform other logic, so I update the chat example to handle JSON data.

Summary of changes:

client.go: 
- Add `Message` struct type.
- Use `ReadJSON` instead of `ReadMessage`.
- Update message trimming logic to handle message body strings instead of byte arrays

hub.go:
- Convert broadcast channel from byte array to `Message` struct type
- `JSON.Marshal` message struct before sending to client

home.html:
- Send stringified json message on form submit
- Parse `onmessage` event data into json

### DIFF
--- a/examples/chat/README.md
+++ b/examples/chat/README.md
@@ -28,32 +28,42 @@ instance of the `Hub` type. The `Hub` maintains a set of registered clients and
 broadcasts messages to the clients.
 
 The application runs one goroutine for the `Hub` and two goroutines for each
-`Client`. The goroutines communicate with each other using channels. The `Hub`
-has channels for registering clients, unregistering clients and broadcasting
-messages. A `Client` has a buffered channel of outbound messages. One of the
-client's goroutines reads messages from this channel and writes the messages to
-the websocket. The other client goroutine reads json messages from the websocket 
-and sends them to the hub.
+`Client`. The application ensures that there is at most one reader and writer on
+a connection by executing all reads and writes for a client on a separate goroutine. 
+This is because Gorilla websocket connections support at most one concurrent reader
+and one concurrent writer. The goroutines communicate with each other using channels. 
+The `Hub` has channels for registering clients, unregistering clients, and broadcasting
+messages. 
+
+A `Client` has a buffered channel of outbound messages which accepts messages of type 
+`Message`. The `Message` type is a struct that represents a JSON message sent by a client. 
+It contains one field representing the `Body` of the message. The Gorilla websocket library
+does not enforce the datatype used to communicate between the client and the server. 
+However, JSON is one of the most commonly used data-interchange formats and was chosen
+for the purposes of this example.
 
 ### Hub 
 
 The code for the `Hub` type is in
 [hub.go](https://github.com/gorilla/websocket/blob/master/examples/chat/hub.go). 
 The application's `main` function starts the hub's `run` method as a goroutine.
-Clients send requests to the hub using the `register`, `unregister` and
+Clients send requests to the hub using the `register`, `unregister`, and
 `broadcast` channels.
 
 The hub registers clients by adding the client pointer as a key in the
 `clients` map. The map value is always true.
 
 The unregister code is a little more complicated. In addition to deleting the
-client pointer from the `clients` map, the hub closes the clients's `send`
+client pointer from the `clients` map, the hub closes the client's `send`
 channel to signal the client that no more messages will be sent to the client.
 
-The hub handles messages by looping over the registered clients and sending the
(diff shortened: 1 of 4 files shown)


examples/chat/client.go
```diff
@@ -104,7 +114,7 @@ func (c *Client) writePump() {
 			// Add queued chat messages to the current websocket message.
 			n := len(c.send)
 			for i := 0; i < n; i++ {
-				w.Write(newline)
+				w.Write([]byte(newline))
```
AlbiDalbi, line 117: `w.Write([]byte('\n'))` is easier to understand.

examples/chat/home.html
```diff
@@ -36,7 +36,7 @@
             appendLog(item);
         };
         conn.onmessage = function (evt) {
-            var messages = evt.data.split('\n');
+            var messages = JSON.parse(evt.data).body.split('\n');
```
AlbiDalbi, line 39: Use JSON array for multiple chat messages instead of stuffing multiple messages in message.body.

    var messages = JSON.parse(evt.Data)
    ...

examples/chat/hub.go
```diff
@@ -40,9 +46,13 @@ func (h *Hub) run() {
 				close(client.send)
 			}
 		case message := <-h.broadcast:
+			data, err := json.Marshal(message)
+			if err != nil {
+				log.Print(fmt.Sprint("Error while marshalling message:", err))
```
AlbiDalbi, line 51: The application blunders ahead after the error.  Error handing is more than logging a message.

elithrar:
My overall feedback is to keep things really simple here - we shouldn't over-optimize or add complexity to the chat example.

A separate, JSON-based example may make more sense.

ibraheemdev:
@elithrar I don't see this as complicating things. Another example with the few changes I committed would probably be repetitive. But if you don't want to merge it, that's fine.

examples/chat/README.md
```diff
@@ -51,9 +51,9 @@ client pointer from the `clients` map, the hub closes the clients's `send`
 channel to signal the client that no more messages will be sent to the client.
 
 The hub handles messages by looping over the registered clients and sending the
-message to the client's `send` channel. If the client's `send` buffer is full,
-then the hub assumes that the client is dead or stuck. In this case, the hub
-unregisters the client and closes the websocket.
+json marshalled message to the client's `send` channel. If the client's `send`
```
elithrar, line 54: Can you expand the docs a bit more? I'm happy to merge this, but this is a good opportunity to improve the docs for newbies. 

* explain the Message type more
* reinforce the single-writer / no concurrent writer requirement at all points (very common mistake)
* make it clear that JSON is an option, and used for the example as it represents a common serialization for WebSocket messages

ibraheemdev: Sure, I can do this. I'm not sure what to say about the no concurrent writer requirement though.

ibraheemdev: I'll try to have it done this week

ibraheemdev: @elithrar I update the documentation in [the latest commit](https://github.com/gorilla/websocket/pull/621/commits/d18f90cf8f1b6ef5cb40d666aacff295d8a1c594).

TaylorRex:
- The application assumes that the JSON sent to the client does not contain`\n` bytes. This is fragile.  The application will break if the developer replaces json.Marshal with json.MarshalIndent.

- It's not obvious that a c2s websocket message is a single chat message and that a s2c websocket message is newline separated sequence of chat messages.  Issue #639 is an example where a developer copied the example code without understanding this point.  This subtlety in the example can be improved by changing the s2c websocket message to a JSON array of chat messages.  This modification also addresses the previous point.

```
w, err := c.conn.NextWriter(websocket.TextMessage)
if err != nil {
    return
}
io.WriteString(w, "[")
w.Write(message)
// Add queued chat messages to the current websocket message.
n := len(c.send)
for i := 0; i < n; i++ {
    io.WriteString(w, ",")
    w.Write(<-c.send)
}
io.WriteString(w, "]")

...

var messages = JSON.parse(evt.data)

```

- With the switch to JSON, there's no need to replace `'\n'`with `' '` in the incoming chat messages.  Delete that code that code to simplify the example.

- There are two kinds of messages at play in the example: websocket messages and chat messages.  Rename `Message` to `ChatMessage` to help make that distinction clear.

- There is a minuscule performance benefit for the `space` and `newline` byte slice variables.  There is no performance benefit for the string versions of these variables. Delete the variables and use the string literals directly.

elithrar:
Timing this one out.

PR #634, opened 2020-09-19
Fix example scroll logic judgement bug

li1234yun:
Fixes #

**Summary of Changes**

1. Fix example/command/home.html appendLog() function `page scroll logic judgement always is true` bug

### DIFF
--- a/examples/command/home.html
+++ b/examples/command/home.html
@@ -9,10 +9,10 @@ window.onload = function () {
     var log = document.getElementById("log");
 
     function appendLog(item) {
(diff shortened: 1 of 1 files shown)


TaylorRex:
The PR changes how scrolling works in the UI.

Old:  The log scrolls new messages into view only when the log is scrolled to the bottom. This allows users to scroll back to read old messages without interruption from new messages.

New: Log scrolls to bottom on all new messages.

MonicaRJones:
If this PR is accepted, the examples/chat/README.md file should be updated to describe the new scrolling UI.

garyburd:
Closed per comments from ghosts.

PR #638, opened 2020-09-27
make SecWebsocketVersion const

### LINKED ISSUE #637: [enhancement]make Sec-Websocket-Version const
## what I sugget
make Sec-Websocket-Version const

### suggested Code place
server.go

```
if !tokenListContainsValue(r.Header, "Sec-Websocket-Version", "13") {
  return u.returnError(w, r, http.StatusBadRequest, "websocket: unsupported version: 13 not found in 'Sec-Websocket-Version' header")
}
```
### reason
I think making webSocket version const is better because version should be updated.
In order to readable for developers and make it easy to change version of web-socket, I suggest about this.

### Pull Request
https://github.com/gorilla/websocket/pull/638

Thanks in advance.

Hiroya3:
Fixes #637

**Summary of Changes**

1. make SecWebsocketVersion const

### DIFF
--- a/server.go
+++ b/server.go
@@ -122,6 +122,7 @@ func (u *Upgrader) selectSubprotocol(r *http.Request, responseHeader http.Header
 // response.
 func (u *Upgrader) Upgrade(w http.ResponseWriter, r *http.Request, responseHeader http.Header) (*Conn, error) {
 	const badHandshake = "websocket: the client is not using the websocket protocol: "
+	const SecWebsocketVersion = "13"
 
 	if !tokenListContainsValue(r.Header, "Connection", "upgrade") {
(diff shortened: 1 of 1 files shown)


TaylorRex:
Is there a reason for exporting the `SecWebsocketVersion` constant?

See issue for other comments on the proposed change.

Hiroya3:
>Is there a reason for exporting the SecWebsocketVersion constant?

I thought SecWebsocketVersion is changeable.
Exporting constant is helpful  for developer not to mind SecWebsocketVersion, this means SecWebsocketVersion is following RFC6455.

As https://tools.ietf.org/html/rfc6455, version is now 13.
But if we have to change follow by RFC6455, we change only version of const.

Hiroya3:
As I wrote https://github.com/gorilla/websocket/issues/637, 
my suggestion will not enhance this project because what developer have to understand will be added in case of using const only in one place.
I will close this PR.

PR #671, opened 2021-03-14
improve echo example

hellflame:
for better example

1. output element height resize to 70vh
2. scroll to bottom on every 'print'


elithrar (approved):
Thanks for this: it seems like a small but worthwhile change.

PR #680, opened 2021-04-02
Include buffered data when returning underlying connection (#679)

### LINKED ISSUE #679: [bug] Conn.UnderlyingConn can return partially-read connection
`Client.DialContext` uses buffered reader to read from the underlying network connection. It's possible that some response data will get buffered along with the headers. We've have observed this happening on slow connection. 

Calling `Conn.UnderlyingConn` returns connection where some data has been read into `Conn.br` without a way to access it. This is why `http.Hijacker.Hijack()` function returns `bufio.ReadWriter` along with the connection.

ipostelnik:
Fixes #679

When returning underlying connection and there's buffered data, return a wrapper that gives access to buffered data first.

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -1160,9 +1160,32 @@ func (c *Conn) SetPongHandler(h func(appData string) error) {
 	c.handlePong = h
 }
 
+// unbufferConn wraps underlying connection to give readers access to buffered unprocessed data
+type unbufferConn struct {
+	net.Conn
+	br *bufio.Reader
+}
+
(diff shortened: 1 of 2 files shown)


MonicaRJones:
This PR breaks applications that use `UnderlyingConn()` to access the `*net.TCPConn` or `*tls.Conn` directly.    An example use is [setting the keep alive period](https://godocs.io/net#TCPConn.SetKeepAlivePeriod).

ipostelnik:
This is better handled by adding `Hijack` method that doesn't break existing users of `UnderlyingConn`

PR #686, opened 2021-05-07
Fix examples / command: kill runned process onclose

lynxzp:
After closing browser tab, or refresh tab, or when example is shooting down, child process may still be running. So I suggest to add defer proc.Kill()

It can happen when example running long time working program, or interactive program.

Fixes #

**Summary of Changes**

1. add `defer proc.Kill()` after `os.StartProcess(...)`

### DIFF
--- a/examples/command/main.go
+++ b/examples/command/main.go
@@ -132,6 +132,7 @@ func serveWs(w http.ResponseWriter, r *http.Request) {
 		internalError(ws, "start:", err)
 		return
 	}
+	defer proc.Kill()
 
 	inr.Close()
 	outw.Close()


MonicaRJones:
The current function  has code to kill a stubborn process. Does that code not work?

The call to proc.Wait at the end of the function waits on the process. There is no process to kill by the time the proposed new call to proc.Kill executes.

lynxzp:
Looks like you are right, and proc.Kill() can't do anything. But I'm sure that it helped me yesterday.
Also yesterday I updated system and rebooted it. Now proc.Kill() not help anymore. **So my patch is not actual.**

**But issue still present** - child process is running after closing function.

proc.Wait() work for regular process, like bash, but not in my case.

I have some kind of special process: it's binary that run tree of processes. It also run telnet and redirect stdin/stdout to telnet. Excuse me, can't share this is financial proprietary software.

At the end of the function telnet is killed, but not tree of processes.

Also when I kill golang process, child tree is closed too. I don't know how it works.

MonicaRJones:
@lynxzp This PR should be closed or updated to code that does something.

lynxzp:
Of course. I would like to create issue. But can't reproduce on another software. And it's just an example in your library. So maybe in future when I find out something.

PR #688, opened 2021-05-13
Support non-blocking read 

### LINKED ISSUE #481: Support evented read
This feature will avoid the need to dedicate a goroutine to read on the connection and enable read buffer pooling.

See [eranyanay/1m-go-websockets](https://github.com/eranyanay/1m-go-websockets) for memory reduction possible with this feature.

See [Go #15735](https://golang.or.g/issue/15735) for discussion about network connection read events.

iluozhiqiang:
Fixes #481 

In the case of massive connections, we hope to use a single goroutine to cooperate with epoll to realize the message reading of the connection, instead of starting a goroutine separately for each connection.
Because goroutines themselves also have overhead, the scheduling performance of goroutines will also decrease linearly when there are massive connections.

**Summary of Changes**

1. add allowReadControlMessages field on Conn
2. add AllowReadControlMessages method on Conn, the method set true of allowReadControlMessages on Conn
3. Change NextReader method , if  allowReadControlMessages is true then return control message

### DIFF
--- a/examples/echo/server.go
+++ b/examples/echo/server.go
@@ -69,6 +69,7 @@ window.addEventListener("load", function(evt) {
         var d = document.createElement("div");
         d.textContent = message;
         output.appendChild(d);
+        output.scroll(0, output.scrollHeight);
     };
 
     document.getElementById("open").onclick = function(evt) {
@@ -126,7 +127,7 @@ You can change the message and send multiple times.
 <button id="send">Send</button>
 </form>
 </td><td valign="top" width="50%">
-<div id="output"></div>
+<div id="output" style="max-height: 70vh;overflow-y: scroll;"></div>
 </td></tr></table>
 </body>
 </html>


MonicaRJones:
This PR breaks out of the loop in NextReader after processing control frames.  

NextReader returns `noframe, nil, nil` for PING and PONG messages.  Because ReadMessage and ReadJSON expect a non-nil reader on success, those functions will panic on a control message.  

The PR does not handle the case where data remains in the bufio.Reader.  If two messages were slurped up in to the bufio.Reader, the second message will not be handled until a third messages is received.   The third message will not be received until the fourth message and so on.

Given the above, the feature should be implemented something like this:

```
if c.allowReadControlMessages  && c.bf.Len() == 0 {
     return noframe, nil, ErrNoData
}
```

where `ErrNoData` is a new exported error value.

AllowReadControlMessages is not a good name for the public API.  Control messages are always allowed.  The name should somehow indicate that the feature is about returning from NextReader on no data.

The function that enables the feature should take a bool argument so that the application can turn the feature on and off.

MonicaRJones:
A read buffer pool is the next obvious change for the non-blocking scenario.  Any PR to support the non-blocking scenario should be a complete solution.  We should avoid littering the API with partial attempts at fixing the problem. 

There is [active discussion](https://github.com/golang/go/issues/15735) on non-blocking read in the standard library.   The standard library solution for non-blocking read may have an impact on how the feature is exposed here.  If network connections invoke a callback when data is ready, this package should probably do the same.  On the other hand,  nothing has happened in the five years since the issue was opened.

This PR is missing tests.  All features should have a test.  Hold off on that until the other issues are addressed.

PR #697, opened 2021-06-11
fix bug in example

lirizhong97:
Fixes #

**Summary of Changes**

1. fix bug in example chat, which adjust the position of invoking function SetReadDeadline.

### DIFF
--- a/examples/chat/client.go
+++ b/examples/chat/client.go
@@ -59,9 +59,9 @@ func (c *Client) readPump() {
 		c.conn.Close()
 	}()
 	c.conn.SetReadLimit(maxMessageSize)
-	c.conn.SetReadDeadline(time.Now().Add(pongWait))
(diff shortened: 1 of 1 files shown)


MonicaRJones:
Setting the deadline before the loop and in the pong handler should be sufficient.  

Can you describe a scenario where the current code does not work?

lirizhong97:
The deadlines are fixed points in (wallclock) time. Unlike timeouts, they don’t reset after a new activity. Each activity on the connection must therefore set a new deadline. How do you think?

lirizhong97:
And, please check the go document for explaination:
// SetReadDeadline sets the deadline for future Read calls
// and any currently-blocked Read call.
// A zero value for t means Read will not time out.
SetReadDeadline(t time.Time) error

MonicaRJones:
The deadline is extended on each PONG message.  The program does not assume that the deadline is a timeout.

What is the bug fixed by this PR?  Did you observe a problem?  Do you have a way to reproduce the problem?

PR #715, opened 2021-08-12
Static analysis : gosec high issue : G402 : TLS MinVersion too low.

Momotoculteur:
# Issue: 
**$ for mod in $(find . -name go.mod); do gosec -quiet -confidence high -severity high $(dirname $mod) || rc=1 ; done**
Results:
[github.com/gorilla/websocket/client_clone.go:13] - G402 (CWE-295): TLS MinVersion too low. (Confidence: HIGH, Severity: HIGH)
    12: 	if cfg == nil {
   13: 		return &tls.Config{}
    14: 	}
Summary:
  Gosec  : dev
  Files  : 15
  Lines  : 3475
  Nosec  : 0
  Issues : 1

### DIFF
--- a/client_clone.go
+++ b/client_clone.go
@@ -10,7 +10,9 @@ import "crypto/tls"
 
 func cloneTLSConfig(cfg *tls.Config) *tls.Config {
 	if cfg == nil {
-		return &tls.Config{}
+		return &tls.Config{
+			MinVersion: tls.VersionTLS12
+		}
 	}
 	return cfg.Clone()
 }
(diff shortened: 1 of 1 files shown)


elithrar:
In the client example, we shouldn’t be implicitly choosing the `minVersion` for a user behind the scenes.

Users can supply a tls.Config that sets the minimum version for the application, which will be correctly cloned. 

The gosec linter is for the _application_ author in this case, not the library. Although we should strive to have good, safe defaults, I am reticent to make a change that is effectively a breaking change to older clients.

MonicaRJones:
The net/http package  does not set the min version when [cloning a nil TLS config](https://github.com/golang/go/blob/c85695a117f1ec3b800ba14570876cfcd2075c1f/src/net/http/transport.go#L2858).

This package should follow the net/http package. 

@Momotoculteur  Are there plans to change the net/http package?

garyburd:
Closed per comments from elithrar and a ghost.

PR #730, opened 2021-10-18
fixed to use context.Context in TLS handshake

Code-Hex:
Relate to https://github.com/golang/go/issues/32406

**Summary of Changes**

- I fixed to allow canceling of timeouts and other events using context during TLS Handshake.
  - https://pkg.go.dev/crypto/tls#Conn.HandshakeContext
- As the link says, Go Team has a migration plan for [build constraints](https://tip.golang.org/doc/go1.17#:~:text=mod%20download%20all.-,//go%3Abuild%20lines,-The%20go%20command). So I fixed it so that it is supported by auto-generation.

### DIFF
--- a/client.go
+++ b/client.go
@@ -314,9 +314,9 @@ func (d *Dialer) DialContext(ctx context.Context, urlStr string, requestHeader h
 
 		var err error
 		if trace != nil {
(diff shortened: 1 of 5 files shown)


garyburd:
Thank you for your contribution. I used your changes in #751.

Issue #745, opened 2021-12-09
[feature] Dialer: Support for custom NetDial for TLS


lluiscampos:
**Is your feature request related to a problem? Please describe.**
> A clear and concise description of what the problem is - e.g. "I'm always frustrated when [...]"

On the `Dialer`, I need to use a custom `NetDial` method for TLS connections to be able to use some advance authentication methods like Mutual TLS using a client key from a Hardware Security Module (HSM). I have some go code wrapping OpenSSL that does the TLS dialing, including the handshake.

This works fine for regular `https` requests by setting this dial function as `DialTLS` for `net/http.Transport`.

The problem in `gorilla/websocket` is that the current `Dialer` interface only supports setting one pair of dial methods (`NetDial` and `NetDialContext`) which is used both for `http` and `https` connections. Later, for `https` a handshake is performed using `TLSClientConfig`, but in my case `tls.Config` fields are not enough to do the handshake, code [here](https://github.com/gorilla/websocket/blob/e8629af678b7fe13f35dff5e197de93b4148a909/client.go#L307).

**Describe the solution you'd like**
>What would the feature look like? How would it work? How would it change the API?

My suggestion is to have two pair of dial functions, `NetDial` + `NetDialContext`, and `NetDialTLS` + `NetDialTLSContext`. The latter pair would be used on `wss`/`https` connections. To avoid unexpected behavior when setting `TLSClientConfig`, if either `NetDialTLS` or `NetDialTLSContext` are set the code would assume that the TLS handshake has been performed there and `TLSClientConfig` can be ignored.

This API change would be completely backwards compatible, and it will better mimic the API of `net/http.Transport` (doc [here](https://pkg.go.dev/net/http#Transport)).

I do have a working code for exact this API change that I am suggesting. I want to add some tests before submitting a PR. In the meanwhile, any feedback is welcome :smiley: 

**Describe alternatives you've considered**
> Are there alternatives you've tried, and/or workarounds in-place?

My workaround was to force the url to passed to `Dial` be `ws`, so that [this specific block](https://github.com/gorilla/websocket/blob/e8629af678b7fe13f35dff5e197de93b4148a909/client.go#L307) is skip... However this only works on certain circumstances and it is very hacky.

lluiscampos:
> **Describe the solution you'd like**
> 
> > What would the feature look like? How would it work? How would it change the API?
> 
[...]

PR #746, opened 2021-12-10
Dialer: add optional method NetDialTLSContext

lluiscampos:
Fixes issue: https://github.com/gorilla/websocket/issues/745

With the previous interface, NetDial and NetDialContext were used for
both TLS and non-TLS TCP connections, and afterwards TLSClientConfig was
used to do the TLS handshake.

While this API works for most cases, it prevents from using more advance
authentication methods during the TLS handshake, as this is out of the
control of the user.

This commits introduces another a new dial method, NetDialTLSContext,
which is used when dialing for TLS/TCP. The code then assumes that the
handshake is done there and TLSClientConfig is not used.

This API change is fully backwards compatible and it better aligns with
net/http.Transport API, which has these two dial flavors. See:
https://pkg.go.dev/net/http#Transport

### DIFF
--- a/client.go
+++ b/client.go
@@ -56,8 +56,14 @@ type Dialer struct {
 	NetDial func(network, addr string) (net.Conn, error)
 
 	// NetDialContext specifies the dial function for creating TCP connections. If
-	// NetDialContext is nil, net.DialContext is used.
+	// NetDialContext is nil, NetDial is used.
 	NetDialContext func(ctx context.Context, network, addr string) (net.Conn, error)
+
+	// NetDialTLSContext specifies the dial function for creating TLS/TCP connections. If
+	// NetDialTLSContext is nil, NetDialContext is used.
+	// If NetDialTLSContext is set, Dial assumes the TLS handshake is done there and
+	// TLSClientConfig is ignored.
+	NetDialTLSContext func(ctx context.Context, network, addr string) (net.Conn, error)
 
 	// Proxy specifies a function to return a proxy for a given
 	// Request. If the function returns a non-nil error, the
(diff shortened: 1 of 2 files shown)


lluiscampos:
The first commit is unrelated to the issue at hand, but otherwise the CI was failing on `gofmt` for latest Go version 1.17.

garyburd:
- Remove DialTLS field.  The corresponding field in net/http Transport is deprecated, so there's no need to add it here.
- Convert map of tests to slice of tests for consistency with other tests.

```
	tests := []struct {
                name              string 
		server            *httptest.Server // server to use
		netDial           func(network, addr string) (net.Conn, error)
		netDialContext    func(ctx context.Context, network, addr string) (net.Conn, error)
		netDialTLS        func(network, addr string) (net.Conn, error)
		netDialTLSContext func(ctx context.Context, network, addr string) (net.Conn, error)
		tlsClientConfig   *tls.Config
	}{
		 name: "HTTP server, all NetDial* defined, shall use NetDialContext",
	         server: server,
                 …
		
```

lluiscampos:
Thanks @garyburd for your feedback. I updated now the pull request following your suggestion. I updated the commit message accordingly too.

> * Remove DialTLS field.  The corresponding field in net/http Transport is deprecated, so there's no need to add it here.

As `Dialer` already had `Dial` and `DialContext` I thought it was good to add the two... but being this a new field, makes sense not to add the deprecated one.

> 
>     * Convert map of tests to slice of tests for consistency with other tests.

Roger that :+1:

garyburd:
Thank you for your contribution.  👍 for the included test.

Issue #617, opened 2020-08-03
[bug] Sec-WebSocket-Key header accepting longer values than RFC6455 accepts.


andrewhodel:
I sent a 20 byte value as Sec-WebSocket-Key and everything works fine, but the RFC states that the value must be exactly 16 bytes when decoded from base64!

That's a problem because when I try to use the client I am writing and testing against this websocket server, it fails on another websocket server.

I'm letting you know so we can all follow the RFC.

**Describe the bug**
> A clear and concise description of what the bug is.

…

**Versions**
> Go version: `go version`
> package version: run `git rev-parse HEAD` inside the repo

…

**Steps to Reproduce**
> How can the bug be triggered?

…

**Expected behavior**
> What output or behaviour were you expecting instead?

…

**Code Snippets**
> A minimum viable code snippet can be useful! (use backticks to format it).

…

PR #752, opened 2022-01-02
Add check for Sec-WebSocket-Key header

hirasawayuki:
Fixes issue: #617

Added the following two fixes
- Check that Sec-websocket-key is encoded as base64
- Check that Sec-Websocket-Key is encoded from 16 bytes long

[RFC6455](https://datatracker.ietf.org/doc/html/rfc6455#section-4.2.1) states the following:
> A |Sec-WebSocket-Key| header field with a base64-encoded value that, when decoded, is 16 bytes in length.

### DIFF
--- a/server.go
+++ b/server.go
@@ -154,8 +154,8 @@ func (u *Upgrader) Upgrade(w http.ResponseWriter, r *http.Request, responseHeade
 	}
 
 	challengeKey := r.Header.Get("Sec-Websocket-Key")
-	if challengeKey == "" {
-		return u.returnError(w, r, http.StatusBadRequest, "websocket: not a websocket handshake: 'Sec-WebSocket-Key' header is missing or blank")
+	if !isValidChallengeKey(challengeKey) {
+		return u.returnError(w, r, http.StatusBadRequest, "websocket: not a websocket handshake: 'Sec-WebSocket-Key' header must be Base64 encoded value of 16-byte in length")
 	}
 
 	subprotocol := u.selectSubprotocol(r, responseHeader)
--- a/util.go
+++ b/util.go
@@ -281,3 +281,18 @@ headers:
 	}
 	return result
 }
+
+// isValidChallengeKey checks if the argument meets RFC6455 specification.
+func isValidChallengeKey(s string) bool {
+	// From RFC6455:
+	//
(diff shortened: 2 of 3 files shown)


garyburd:
Do popular libraries for other programming languages reject invalid keys?  I don't want this package to the only library that rejects invalid keys. 

Assuming that it's common practice to reject invalid keys, please move the logic to new function `isValidChallengeKey(s string) bool { }` in  [util.go](https://github.com/gorilla/websocket/blob/2d6ee4c55cc9e9dc0eb5929f32a999213e25256f/util.go#L33).  I want to keep the functionality related to keys in one file. 

It's OK to return a single error message for invalid keys.

hirasawayuki:
@garyburd 
Thanks for the review! I checked the [ws(nodejs library)](https://github.com/websockets/ws), and it seems to return HTTP Status 400 if it doesn't match `/^[+/0-9A-Za-z]{22}==$/`. 
So I saw no problem in rejecting an invalid key.

https://github.com/websockets/ws/blob/5edf1f4a1b1750109c1bb56eff7ad78902eee7dc/lib/websocket-server.js#L18
https://github.com/websockets/ws/blob/5edf1f4a1b1750109c1bb56eff7ad78902eee7dc/lib/websocket-server.js#L236-L245

garyburd:
LGTM. 

The nodejs library provides strong evidence that the check is OK, but I want check more libraries before accepting the PR.  One cause for concern is that the  [Autobahn|Testsuite](https://github.com/crossbario/autobahn-testsuite)  does not test invalid keys.

plasticsax:
Simplify: `return err == nil && len(decoded) == 16`

hirasawayuki:
@garyburd
I'll let you be the judge! 😃 

@plasticsax
Thankyou! I fixed it.(3ae5d9c)

andrewhodel:
There is a maximum limit of the incoming data per the base64 limit. 

You should add an error from the upgrader if the size is exceeded for two reasons:

1. The response isn't unknowingly handled in the module code like go's http 301 when there is a .. in the url path. 

2. The developer gets to decide how strict to be instead of a lower level structure that doesn't participate in these cost decisions.

andrewhodel:
The base64 alphabet has a limit per byte with regards to representation size.

PR #761, opened 2022-01-09
Correct way to save memory using write buffer pool and freeing net.http default buffers

FMLS:
**Summary of Changes**

1. Add an example that uses the write buffer pool

The loop process of the websocket connection is inner the http handler at existing examples， This usage will cause the 8k buffer(4k read buffer + 4k write buffer) allocated  by net.http can't be GC(Observed by heap profiling, see picture below) . The purpose of saving memory is not achieved even if the WriteBufferPool is used. 

In example bufferpool, server process websocket connection in a new goroutine, and the goroutine created by the net.http will exit, then the 8k buffer will be GC.

heap

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,6 +20,7 @@ Gorilla WebSocket is a [Go](http://golang.org/) implementation of the
 * [Command example](https://github.com/gorilla/websocket/tree/master/examples/command)
 * [Client and server example](https://github.com/gorilla/websocket/tree/master/examples/echo)
 * [File watch example](https://github.com/gorilla/websocket/tree/master/examples/filewatch)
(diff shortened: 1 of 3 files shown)


wraix:
Maybe the ideas posted here could be reflected to get even better at large volume connection handling or at least give the options to?

[A Million WebSockets and Go](https://www.freecodecamp.org/news/million-websockets-and-go-cc58418460bb/)

FMLS:
> Maybe the ideas posted here could be reflected to get even better at large volume connection handling or at least give the options to?
> 
> [A Million WebSockets and Go](https://www.freecodecamp.org/news/million-websockets-and-go-cc58418460bb/)

yes, it's a good idea in this blog, but golang did not provide a suitable mechanism now
see https://github.com/golang/go/issues/15735

PR #763, opened 2022-01-24
fix #705

### LINKED ISSUE #705: WriteMessage and WriteControl do not check error from SetWriteDeadline [bug]
The `SetWriteDeadline` method from `net.Conn` can return an error. https://golang.org/pkg/net/#Conn

However, this library neglects to check it.
https://github.com/gorilla/websocket/blob/b65e62901fc1c0d968042419e74789f6af455eb9/conn.go#L389
https://github.com/gorilla/websocket/blob/b65e62901fc1c0d968042419e74789f6af455eb9/conn.go#L456

kevinlincg:
Remove writeDealine , use net/Conn SetWriteDeadline and handle error to user.

Fixes #705 

**Summary of Changes**

1. Remove writeDeadline
2. Let SetWriteDeadline is wrap net/Conn SetWriteDeadline function

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -245,13 +245,12 @@ type Conn struct {
 	subprotocol string
 
 	// Write fields
-	mu            chan struct{} // used as mutex to protect write to conn
-	writeBuf      []byte        // frame is constructed in this buffer.
-	writePool     BufferPool
-	writeBufSize  int
-	writeDeadline time.Time
-	writer        io.WriteCloser // the current writer returned to the application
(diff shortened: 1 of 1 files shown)


garyburd:
A call to [WriteControl](https://pkg.go.dev/github.com/gorilla/websocket#Conn.WriteControl) must not overwrite the deadline set by SetWriteDeadline.   Fix is `if err := c.conn.SetWriteDeadline(deadline); err != nil { return err }` in several places.

PR #740, opened 2021-12-03
Tls proxy

philipatl:
Fixes #739 

**Summary of Changes**

1. call `proxy_RegisterDialerType` for the `https` scheme to allow communication to https proxies
2. some scheme-specific setup is necessary in Dialer's DialContext in order to use the TLSClientConfig
3. add a `UsesTLS` method to the `proxy_Dialer` interface to ensure the https forwardDialer can inspect that https is supported, and to adjust if necessary

> PS: Make sure your PR includes/updates tests! If you need help with this part, just ask!

PR #782, opened 2022-05-12
Add optional method ProxyTLSConnection (closes #779)

### LINKED ISSUE #779: [bug] Can't establish websocket connection when using `NetDialTLSContext` and `Proxy` together
**Describe the bug**
I can't connect to a websocket server over a  HTTP proxy when I set both `NetDialTLSContext` and `Proxy` fields on the `websocket.Dialer` struct.  See the code snippet below for more details, it's best explained by code. 

**Versions**
Go version: `go1.18.1 linux/amd64`.
Package version: `v1.5.0`.

**Steps to Reproduce**
1. Get a HTTP proxy to connect to. You could install a local HTTP debugging proxy like Burp Suite, Charles Proxy, MITMProxy, ... on your development machine for testing.
2. Execute the code snippet below and observe that no websocket connection can be established over the proxy you configured.

**Expected behavior**
The websocket dialer should be able to establish a websocket connection over a HTTP proxy while also applying a custom TLS connection without issues.

**Code Snippets**
The following code results in an error like `websocket: bad handshake`  or ` unexpected EOF`  (depends on the proxy):

```go
package main

import (
	"context"
	"crypto/tls"
	"log"
	"net"
	"net/http"
	"net/url"
)

const (
	// Local or remote HTTP proxy.
	proxyUrl = "http://127.0.0.1:8888"

	// Websocket endpoint for testing
	websocketUrl = "wss://echo.websocket.events/"
)

func main() {
	p, _ := url.Parse(proxyUrl)

	dialer := websocket.Dialer{
		Proxy:             http.ProxyURL(p),
		EnableCompression: true,
		NetDialTLSContext: func(ctx context.Context, network, addr string) (net.Conn, error) {
			// TCP dial
			netConn, err := net.Dial(network, addr)
			if err != nil {
				return nil, err
			}

			// 'NetDialTLSContext' also gets called during the proxy CONNECT for some reason (at this point 'network' equals "TCP" and 'addr' equals "127.0.0.1:8888")
			// The HTTP proxy doesn't support HTTPS however, so I return the established TCP connection early.
			// If I don't do this check, the connection hangs forever (tested with several proxies).
			// This feels kinda hacky though, not sure if this is the correct approach...
			if p.Host == addr {
				return netConn, err
			}

			// Example TLS handshake
			tlsConn := tls.Client(netConn, &tls.Config{ServerName: "echo.websocket.events", InsecureSkipVerify: true})
			if err = tlsConn.Handshake(); err != nil {
				return nil, err
			}

			return tlsConn, nil
		},
	}

	conn, _, err := dialer.Dial(websocketUrl, nil)
	if err != nil {
		log.Fatalln(err)
	}

	log.Println(conn.LocalAddr())
	// ...
}
```

When I check the request in burp, it looks like this:

image

As highlighted in red, the host URL became `http://echo.websocket.events:443/`, which obviously should just be `https://echo.websocket.events/`. I've been trying to debug why this is happening exactly all day but can't seem to find the culprit so I need help.

If I remove `NetDialTLSContext` and just specify `TLSClientConfig` everything works as expected. Same result with the `Proxy` field removed. The issue only occurs when both of these fields are set.

Am I missing something obvious?

sleeyax:
Previously, it was impossible to specify both `Proxy` and `NetDialTLSContext` on the websocket `Dialer` without experiencing connection issues. This commit brings a change to the proxy CONNECT flow so that the initial connection is always a normal proxy CONNECT over TCP, while allowing TLS customizations on the existing connection at a later point in time via the new `ProxyTLSConnection` method.

Fixes #779

**Summary of Changes**

Please see my comment on the relevant issue for a detailed explanation and my thought process: https://github.com/gorilla/websocket/issues/779#issuecomment-1125185896

Tests are **not included**, I'd appreciate some help or guidance on that front because I'm unsure what to test, thanks!

### DIFF
--- a/client.go
+++ b/client.go
@@ -65,6 +65,12 @@ type Dialer struct {
 	// TLSClientConfig is ignored.
 	NetDialTLSContext func(ctx context.Context, network, addr string) (net.Conn, error)
 
+	// ProxyTLSConnection specifies the dial function for creating TLS connections through a Proxy. If
+	// ProxyTLSConnection is nil, NetDialTLSContext is used.
+	// If ProxyTLSConnection is set, Dial assumes the TLS handshake is done there and
+	// TLSClientConfig is ignored.
+	ProxyTLSConnection func(ctx context.Context, proxyConn net.Conn) (net.Conn, error)
+
 	// Proxy specifies a function to return a proxy for a given
 	// Request. If the function returns a non-nil error, the
 	// request is aborted with the provided error.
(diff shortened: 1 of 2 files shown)


jaitaiwan:
Thanks for the commit. I know it's been a long time but it would be good to see a test to go with this PR. I'm not fully confident in merging without associated tests.

jaitaiwan:
It also looks like this PR does something similar: https://github.com/gorilla/websocket/pull/740/files maybe consider using that PR to bolster your own.

sleeyax:
> It also looks like this PR does something similar: https://github.com/gorilla/websocket/pull/740/files maybe consider using that PR to bolster your own.

Not sure if this fixes the issue completely. The PR you mention adds support for HTTPS proxies whereas this PR fixes a bug where `NetDialTLSContext` and `Proxy` fields on the `websocket.Dialer` don't work together. See the referenced issues for more details.

sleeyax:
> Thanks for the commit. I know it's been a long time but it would be good to see a test to go with this PR. I'm not fully confident in merging without associated tests.

It's been so long. Can't say when/if I'll be able to find the time to look into this again and add tests.

PR #785, opened 2022-05-22
add WriteText()

Alex41:


amustaque97:
Hi @alex would you mind sharing how this PR is going to help? What problem does it address? Any existing issue? 

Thank you

garyburd:
The value of this method does not outweigh the cost of increasing the API surface area.

PR #791, opened 2022-06-19
Update required versions in CircleCi config

garyburd:


amustaque97:
Quick question on the PR, are we not going to maintain backward compatibility for the removed golang versions? For example version 1.14 or 1.15

PR #818, opened 2022-09-27
Judge error once we get it when calling function WriteJSON

### LINKED ISSUE #817: [feature] Avoid flushing frame twice when calling WriteJSON
**Is your feature request related to a problem? Please describe.**
Currently, the function `WriteJSON` function implementation is shown as the following code snippet.
```go
func (c *Conn) WriteJSON(v interface{}) error {
	w, err := c.NextWriter(TextMessage)
	if err != nil {
		return err
	}
	err1 := json.NewEncoder(w).Encode(v)
	err2 := w.Close()
	if err1 != nil {
		return err1
	}
	return err2
}
```
We will call `w.Close` whatever the `json.NewEncoder(w).Encode(v)` gets an error or not.
The errors raised by the `Encode` method have two types:
- no matter with writing data to the connection.
- error when writing to connection, this is actually raised by the `flushFrame` method.

As a result, if we receive a non-nil error from `json.NewEncoder(w).Encode(v)`, we can return the error directly and no need to flush it twice.

**Describe the solution you'd like**
Return the error directly instead of `flushFrame` twice.
```go
func (c *Conn) WriteJSON(v interface{}) error {
    w, err := c.NextWriter(TextMessage)
    if err != nil {
        return err
    }
    
    if err = json.NewEncoder(w).Encode(v); err != nil {
        return err
    }
    
    return w.Close()
}
```

**Describe alternatives you've considered**
N/A

I'm not sure whether there was a special consideration here. I have searched the issues but found no one. If there are some docs about this, please ping me.
Best regards, thanks.

xieyuschen:
**Summary of Changes**

Check the error once it is raised.

### DIFF
--- a/json.go
+++ b/json.go
@@ -5,15 +5,15 @@
 package websocket
 
 import (
-	"encoding/json"
-	"io"
+    "encoding/json"
+    "io"
 )
 
 // WriteJSON writes the JSON encoding of v as a message.
 //
 // Deprecated: Use c.WriteJSON instead.
 func WriteJSON(c *Conn, v interface{}) error {
-	return c.WriteJSON(v)
+    return c.WriteJSON(v)
 }
 
 // WriteJSON writes the JSON encoding of v as a message.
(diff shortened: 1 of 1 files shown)


xieyuschen:
The details are explained in #817 , could check it:))

Thanks:)

PR #840, opened 2023-07-19
 Update go version & add verification/testing tools

coreydaley:
Fixes #

**Summary of Changes**

1.
2. 
3.

> PS: Make sure your PR includes/updates tests! If you need help with this part, just ask!

PR #846, opened 2023-08-26
Use crypto/rand for mask key

HaloArrow:
This PR fixes issues with a PR in progress:

- Call crypto/rand Reader directly to get four random bytes.
- Use seeded math/rand Reader for reproducible results in tests.


coreydaley:
I have pulled your pull request into https://github.com/gorilla/websocket/pull/840

PR #853, opened 2023-09-28
Update server.go

samyuktha-asokan:
After the connection is upgraded from the HTTP protocol to the WebSocket protocol, the network connection cannot be used for HTTP. The net/http server panics when the application attempts to write an HTTP response to a hijacked connection. You should not use netConn to write the response headers. Instead, you should use the http.ResponseWriter to write the response headers to avoid the log "http: response.WriteHeader on hijacked connection from...."

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Go Version Update
- [ ] Dependency Update

- Related Issue #
- Closes #

## Added/updated tests?

- [ ] Yes
      have not been included_

## Run verifications and test

- [ ] `make test` is passing

### DIFF
--- a/server.go
+++ b/server.go
@@ -266,7 +266,7 @@ func (u *Upgrader) Upgrade(w http.ResponseWriter, r *http.Request, responseHeade
 			return nil, err
 		}
 	}
-	if _, err = netConn.Write(p); err != nil {
+	if _, err = w.Write(p); err != nil {
 		if err := netConn.Close(); err != nil {
 			log.Printf("websocket: failed to close network connection: %v", err)
 		}


pennystevens:
This PR attempts to write the websocket handshake response headers to the connection using ResponseWriter.Write method.  This fails for two reasons: 

- ResponseWriter.Write fails on a hijacked connection.
- Even if ResponseWriter.Write did not fail on the hijacked connection, ResponseWriter.Write writes the data to the response body, not the response header.

Many tests will fail with this change.

> After the connection is upgraded from the HTTP protocol to the WebSocket protocol, the network connection cannot be used for HTTP

This is not true.  The hijacker takes control of the connection and can implement whatever protocol it want, including HTTP.

GreenMarmot:
@jaitaiwan This should be closed. Tests fail with the change.  See previous comment.

AlexVulaj:
I've closed this PR as per the comment above.

PR #854, opened 2023-10-08
Update the Host field in the request header

HaoweiCh:
This update checks for the 'Host' value in the request header, specifically assigning it to 'u.Host' if no value is found. This same logic is applied to 'cfg.ServerName'. The aim of this change is to ensure a correct 'Host' value is used throughout, addressing a bug related to setting the 'Host' field of the header.

- [x] Refactor
- [x] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Go Version Update
- [ ] Dependency Update

- Related Issue #
- Closes #

## Added/updated tests?

- [ ] Yes
      have not been included_

## Run verifications and test

- [x] `make test` is passing

### DIFF
--- a/client.go
+++ b/client.go
@@ -189,6 +189,10 @@ func (d *Dialer) DialContext(ctx context.Context, urlStr string, requestHeader h
 		return nil, nil, errMalformedURL
 	}
 
+	host := requestHeader.Get("Host")
+	if host == "" {
+		host = u.Host
+	}
 	req := &http.Request{
 		Method:     http.MethodGet,
 		URL:        u,
@@ -196,7 +200,7 @@ func (d *Dialer) DialContext(ctx context.Context, urlStr string, requestHeader h
 		ProtoMajor: 1,
 		ProtoMinor: 1,
 		Header:     make(http.Header),
-		Host:       u.Host,
+		Host:       host,
 	}
 	req = req.WithContext(ctx)
 
@@ -347,7 +351,11 @@ func (d *Dialer) DialContext(ctx context.Context, urlStr string, requestHeader h
 
 		cfg := cloneTLSConfig(d.TLSClientConfig)
 		if cfg.ServerName == "" {
-			cfg.ServerName = hostNoPort
+			if host != "" {
+				cfg.ServerName = host
+			} else {
+				cfg.ServerName = hostNoPort
+			}
 		}
 		tlsConn := tls.Client(netConn, cfg)
 		netConn = tlsConn


pennystevens:
What is the bug that this fixes?  What is the scenario where the application specifies the host in the headers instead of the URL?

HaoweiCh:
When it comes to dealing with dynamic servers that the application controls, the ability to send data through a Content Delivery Network (CDN) is a big deal for me. It becomes especially important when I have to forward WebSocket connections. This feature allows me to use DNS over HTTPS (DoH) to look up and specifically choose the server's IP address. Particularly with Cloudflare CDN, I can even customize and select specific IP addresses for Cloudflare's edge nodes. To put it simply, this gives me the flexibility and control I need to ensure smooth communication between the application and the server.

pennystevens:
Is this a correct statement of the problem: specify the sever address separate from the host used in the request header and when negotiating TLS.

If the statement of the problem is correct, then use the [net dial](https://pkg.go.dev/github.com/gorilla/websocket#Dialer.NetDialContext) function:

     addr := "example1.com:80" // server address
     url := "wss://example2.com/ws"  // use example2.com in host header and for TLS
   
     wsd := websocket.Dialer{
         NetDialContext: func(ctx context.Context, network, urlAddr string) (net.Conn, error) {
              // urlAddr is the address taken from the request
             // URL. We ignore that address when dialing the 
             // underlying network connection.
             var nd net.Dialer        
             return nd.DialContext(ctx, network, addr)
        },
    }
    ws, _, err := wsd.DialContext(ctx, url, nil)

pennystevens:
This package should follow the design of the net/http package, not curl.  The net/http package ignores the application specified `Host` header ([start here to view the code](https://github.com/golang/go/blob/aa8e4c5e610145af682d48b064fa0d28ed9a91e4/src/net/http/request.go#L97-L103)).

This package solves your underlying problem using the net dial function.  Whether the package should add additional ways to solve the problem is up to the package maintainers.

If the maintainers do decide to proceed forward with this PR, then the PR should be updated to include:
- Documentation for the new feature.
- A test for the new feature.
- A PR comment analyzing what happens to existing applications that set the `Host` header. Does this PR break those applications, or were the applications flawed to begin with?

I am not asking you to do any of the items above. That's up to the maintainers.

jaitaiwan:
I'm sad to see zak seems to be a deleted user. I agree with their comment and so does @canelohill. I don't think it's important for us to include this functionality and there's already example code showing how it can be done by modifying the net dialer. So I'm going to close this one out.

PR #863, opened 2023-11-08
Do not handle network error in `SetCloseHandler()`

nak3:
The 666c197 added an error handling in `SetCloseHandler()` and peer stops getting `CloseError` when network issue like `write: broken pipe` happens because the close handle returns the error.

Hence this patch changes to skip network error handling.

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -1159,8 +1159,12 @@ func (c *Conn) SetCloseHandler(h func(code int, text string) error) {
 		h = func(code int, text string) error {
 			message := FormatCloseMessage(code, "")
 			err := c.WriteControl(CloseMessage, message, time.Now().Add(writeWait))
-			if err != nil && err != ErrCloseSent {
-				return err
+			if err != nil {
+				if _, ok := err.(net.Error); ok {
+					return nil
+				} else if err != ErrCloseSent {
+					return err
+				}
 			}
 			return nil
 		}
--- a/conn.go
+++ b/conn.go
@@ -1158,14 +1158,7 @@ func (c *Conn) SetCloseHandler(h func(code int, text string) error) {
 	if h == nil {
 		h = func(code int, text string) error {
 			message := FormatCloseMessage(code, "")
-			err := c.WriteControl(CloseMessage, message, time.Now().Add(writeWait))
-			if err != nil {
-				if _, ok := err.(net.Error); ok {
-					return nil
-				} else if err != ErrCloseSent {
-					return err
-				}
-			}
+			_ = c.WriteControl(CloseMessage, message, time.Now().Add(writeWait))
 			return nil
 		}
 	}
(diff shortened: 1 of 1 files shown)


nak3:
For more detail, we have an unit test as https://github.com/knative/serving/blob/65ce2aece44a5bc5550b54aef657a82f0a6ce61b/pkg/autoscaler/statserver/server_test.go#L111 but it stops getting `CloseError` on websocket v1.5.1.

pennystevens:
SetPingHandler has the same issue. Related info:

-  https://github.com/gorilla/websocket/pull/847
- https://github.com/gorilla/websocket/pull/840#pullrequestreview-1597152323

coreydaley:
This may have been fixed in https://github.com/gorilla/websocket/pull/865 can you take a look and verify?

nak3:
No, https://github.com/gorilla/websocket/pull/865 does not fix the issue. I looked at and also verified it.

coreydaley:
Then there is a merge conflict with this pull request now that needs to be resolved then.

pennystevens:
With this change and 865, the close handler will in effect ignore all errors.   Consider changing the code to ignore all errors as [as the code did prior to August of this year](https://github.com/gorilla/websocket/blob/76ecc29eff79f0cedf70c530605e486fc32131d1/conn.go#L1139).

nak3:
> With this change and 865, the close handler will in effect ignore all errors.

I think the close handler with this PR will not ignore all errors but will return some error such as `errInvalidControlFrame` and `errWriteTimeout`.

> Consider changing the code to ignore all errors as as the code did prior to August of this year.
> It's more important to return the CloseError to the application than any error returned from writing the control message.

I sort of agree that the close handle should ignore all errors. But I cannot confirm that so this PR ignores network error only for now.

pennystevens:
- `errWriteTimeout` implements the `net.Error` interface and is therefore ignored by this PR.
- `WriteControl(CloseMessage, FormatCloseMessage(code, ""), t)` will not return `errInvalidControlFrame`.  We know that `CloseMessage` is a valid control frame type and that `FormatCloseMessage(code, "")` is a valid control frame payload. 

The websocket connection ignored the error returned  from echoing the close message until the PR in August.  It seems safe to revert back to the original code.

nak3:
Okay, thank you. Updated

nak3:
@coreydaley Could you please take a look?

nak3:
@coreydaley gentle ping.

AlexVulaj:
There was a recent issue brought up about noisy, unactionable error messages that may be related here: https://github.com/gorilla/websocket/pull/878

I'm in favor of taking a similar approach and ignoring the error in this specific case. @coreydaley thoughts?

nak3:
@coreydaley @AlexVulaj Sorry for bothering you. But could you please take a look?

nak3:
@coreydaley @AlexVulaj gentle ping.

dprotaso:
Hey folks - what's the latest with this PR? Knative is hoping to bump to 1.5.1+ but this is blocking us

AlexVulaj (approved):

AlexVulaj:
@nak3 @dprotaso hey, sorry about the delay. Change is merged - thanks for contributing!

PR #865, opened 2023-11-09
conn.go: default close handler should not return ErrCloseSent.

pnx:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Go Version Update
- [ ] Dependency Update

## Description

Noticed a change in default close handler from 1.5.0 to 1.5.1

1.5.1 Now returns the error from `WriteControl` with `CloseMessage` type. However this results in ErrCloseSent returned if the connection initiated the close handshake. This is normally correct as the connection should not be able to write after a close handshake is initiated, except when calling the close handler. in that case nil should be returned so that `advanceFrame()` can return the actual close message.

## Added/updated tests?

- [x] Yes
      have not been included_

## Run verifications and test

- [ ] `make verify` is passing (fails with `GO-2023-2186 Incorrect detection of reserved device names on Windows in path/filepath` that is unrelated to this change)
- [x] `make test` is passing

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -1163,7 +1163,8 @@ func (c *Conn) SetCloseHandler(h func(code int, text string) error) {
 	if h == nil {
 		h = func(code int, text string) error {
 			message := FormatCloseMessage(code, "")
-			if err := c.WriteControl(CloseMessage, message, time.Now().Add(writeWait)); err != nil {
+			err := c.WriteControl(CloseMessage, message, time.Now().Add(writeWait))
+			if err != nil && err != ErrCloseSent {
 				return err
 			}
 			return nil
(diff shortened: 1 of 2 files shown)


pennystevens:
A similar change is required in the ping handler.  Consider this scenario:

1.  PING message is in flight from peer
2. the application writes a CLOSE message
3. the peer echos the CLOSE message
4. the PING message is received, connection responds with PONG
5. the CLOSE message is received, connection echos the CLOSE message.

This PR addresses step 5 where ErrCloseSent is returned instead of the CLOSE message.

Application never gets to step 5 because the PING handler returns ErrCloseSent and this error is returned from NextReader, ReadMessage, etc. 

In summary: When the connection automatically sends a message in response to a received message, ErrCloseSent should be ignored.  

Version 1.5.0 also ignored timeout errors in this scenario.

pnx:
@coreydaley can you take a look at this?

coreydaley (approved):

PR #869, opened 2023-11-21
fix: add comment for the readBufferSize and writeBufferSize

rfyiamcool:
### summary

😅  Every time I use `gorilla/websocket`, I need to look at the source code and find the default size for readBufferSize and writeBufferSize.

I think the default value should be written in the comment.


AlexVulaj (approved):

GreenMarmot:
The comment added in this PR is accidentally correct.  The PR should be reverted or a test should be added to ensure that the claim made by the comment is correct.  Weirdly, the PR does not add a comment for the Dialer where the default is explicitly 4096 in the code.

PR #870, opened 2023-11-21
fix: use atomic.bool to ensure atomic operation of isWriting

rfyiamcool:
### summary

I think `isWriting` as bool type is not thread-safe operation under concurrency. 😁

For example, multiple Goroutines determine whether iswriting is false before writing data, and if `isWriting` is false, write data.  but, the two operations of querying `isWriting` and setting `isWriting` are not atomic.

### example code

`the following scene`

goroutine 1

```go
func (w *messageWriter) flushFrame(final bool, extra []byte) error {
	...
	
	if c.isWriting {
		panic("concurrent write to websocket connection")
	}

 	// run here 😁

	c.isWriting = true

        c.write(....)

        ....
}
```

goroutine 2

```go
func (w *messageWriter) flushFrame(final bool, extra []byte) error {
	...
	if c.isWriting {
		panic("concurrent write to websocket connection")
	}

 	// run here 😁

	c.isWriting = true

        c.write(....)

        ....
}
```

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -15,6 +15,7 @@ import (
 	"strconv"
 	"strings"
 	"sync"
+	"sync/atomic"
 	"time"
 	"unicode/utf8"
 )
@@ -258,7 +259,7 @@ type Conn struct {
 	writeBufSize  int
 	writeDeadline time.Time
 	writer        io.WriteCloser // the current writer returned to the application
-	isWriting     bool           // for best-effort concurrent write detection
+	isWriting     atomic.Bool    // for best-effort concurrent write detection
 
 	writeErrMu sync.Mutex
 	writeErr   error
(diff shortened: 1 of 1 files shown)


FZambia:
Hello, as written in comment for isWriting - it's a best effort protection from concurrent writes. If you have a race here it's a signal you are writing concurrently - thus a signal you are using Gorilla WebSocket API in a wrong way. It does not support concurrent writes. So for me it seems to be working as intended.

pennystevens:
I agree with FZambia's comment.  Here's some more information:

The current code is similar to the check for concurrent writes in Go maps.  The Go map check uses a bit flag, not an atomic bool (see uses of `hashWriting` in [runtime/map.go](https://github.com/golang/go/blob/master/src/runtime/map.go)). 

The [websocket package documentation](https://pkg.go.dev/github.com/gorilla/websocket#hdr-Concurrency) says:

> Connections support one concurrent reader and one concurrent writer.

There is not a race when the application uses the API per the documentation.

rfyiamcool:
> I agree with FZambia's comment. Here's some more information:
> 
> The current code is similar to the check for concurrent writes in Go maps. The Go map check uses a bit flag, not an atomic bool (see uses of `hashWriting` in [runtime/map.go](https://github.com/golang/go/blob/master/src/runtime/map.go)).
> 
> The [websocket package documentation](https://pkg.go.dev/github.com/gorilla/websocket#hdr-Concurrency) says:
> 
> > Connections support one concurrent reader and one concurrent writer.
> 
> There is not a race when the application uses the API per the documentation.

You are right. 😁  @pennystevens 

[
image
](
image
)

PR #871, opened 2023-11-22
Upgrader.Upgrade: use http.ResposnseController

ninedraft:
Since go1.20 [http.ResponseController](https://pkg.go.dev/net/http#ResponseController) is a preferable tool for connection hijacking. It allows to use custom response writer interceptors with `Unwrap()  http.ResponseWriter` method.

### DIFF
--- a/server_test.go
+++ b/server_test.go
@@ -7,8 +7,10 @@ package websocket
 import (
 	"bufio"
 	"bytes"
+	"errors"
 	"net"
 	"net/http"
+	"net/http/httptest"
 	"reflect"
 	"strings"
 	"testing"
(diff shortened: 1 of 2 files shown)


server_test.go
```diff
@@ -147,3 +149,23 @@ func TestBufioReuse(t *testing.T) {
 		}
 	}
 }
+
+func TestHijack_NotSupported(t *testing.T) {
+	t.Parallel()
```
AlexVulaj, line 154: Let's remove this line since we aren't parallelizing tests in this repo.

AlexVulaj (approved):
Thanks for the contribution! Approved.

PR #878, opened 2023-11-29
Remove noisy printf in NextReader() and beginMessage()

### LINKED ISSUE #852: [BUG] Debug messages are printed to log

### Current Behavior

The code has started printing debug messages to via log.Printf(). For example:

`2023/09/19 09:27:23 websocket: discarding reader close error: io: read/write on closed pipe`

It seems to have been introduced recently:

https://github.com/gorilla/websocket/blame/666c197fc9157896b57515c3a3326c3f8c8319fe/conn.go#L1025

### Expected Behavior

No debug messages are printed to the console.

### Steps To Reproduce

_No response_

### Anything else?

_No response_

bcreane:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Go Version Update
- [ ] Dependency Update

## Description
A recent check in causes `NextReader()` and `beginMessage()` to emit noisy, non-actionable messages that fill up application logs. This PR restores the previous approach of ignoring reader/writer close errors. This should address https://github.com/gorilla/websocket/issues/852.

- Related Issue #
- Closes #852 

## Added/updated tests?

- [ ] Yes
- [x] No, and this is why: no tests exist to track spurious log entries, so eliding those entries doesn't require a test update

## Run verifications and test

- [x] `make test` is passing

### DIFF
--- a/compression.go
+++ b/compression.go
@@ -8,7 +8,6 @@ import (
 	"compress/flate"
 	"errors"
 	"io"
-	"log"
 	"strings"
 	"sync"
 )
@@ -135,9 +134,7 @@ func (r *flateReadWrapper) Read(p []byte) (int, error) {
 		// Preemptively place the reader back in the pool. This helps with
 		// scenarios where the application does not call NextReader() soon after
 		// this final read.
-		if err := r.Close(); err != nil {
-			log.Printf("websocket: flateReadWrapper.Close() returned error: %v", err)
-		}
+		_ = r.Close()
 	}
 	return n, err
 }
--- a/conn.go
+++ b/conn.go
@@ -10,7 +10,6 @@ import (
 	"encoding/binary"
 	"errors"
 	"io"
-	"log"
 	"net"
 	"strconv"
 	"strings"
@@ -490,9 +489,7 @@ func (c *Conn) beginMessage(mw *messageWriter, messageType int) error {
 	// probably better to return an error in this situation, but we cannot
 	// change this without breaking existing applications.
 	if c.writer != nil {
-		if err := c.writer.Close(); err != nil {
-			log.Printf("websocket: discarding writer close error: %v", err)
-		}
+		_ = c.writer.Close()
 		c.writer = nil
 	}
 
@@ -1021,9 +1018,7 @@ func (c *Conn) handleProtocolError(message string) error {
 func (c *Conn) NextReader() (messageType int, r io.Reader, err error) {
 	// Close previous reader, only relevant for decompression.
 	if c.reader != nil {
-		if err := c.reader.Close(); err != nil {
-			log.Printf("websocket: discarding reader close error: %v", err)
-		}
+		_ = c.reader.Close()
 		c.reader = nil
 	}
 


bcreane:
I realize that ignoring these errors may not be ideal, however I note that an earlier contributor points out that returning these errors is not feasible w/o breaking the current API contract:
>      // Close previous writer if not already closed by the application. It's
>      // probably better to return an error in this situation, but we cannot
>      // change this without breaking existing applications.

pennystevens:
**Regarding c.reader.Close()** 

The call invokes [messageReader.Close](https://github.com/gorilla/websocket/blob/6f5d2139f4b96d5edfe6c1aac4fe87fed8d7f9fd/conn.go#L1112) or [flateReadWrapper.Close](https://github.com/gorilla/websocket/blob/6f5d2139f4b96d5edfe6c1aac4fe87fed8d7f9fd/compression.go#L145).  

messageReader.Close always returns nil.

flatReadWrapper.Close returns io.ErrClosedPipe when closed previously. This error is safe to ignore.  If the read wrapper is not closed, then flatReadWrapper.Close calls the flat reader [Close](https://github.com/golang/go/blob/858cd8da569938913541d013af27a2a2a875fcb5/src/compress/flate/inflate.go#L355) method.  The error returned from the flat Reader Close method is safe to ignore because that error would have been handled in a previous call to Read.  

Summary: it's safe to ignore the error returned from c.reader.Close().

There is another call to flatReadWrapper.Close [here](https://github.com/gorilla/websocket/blob/6f5d2139f4b96d5edfe6c1aac4fe87fed8d7f9fd/compression.go#L139).  That code should also ignore the returned error.
  
**Regarding c.writer.Close()**

The comment says that returning the error will break existing applications. Prior to the addition of that comment, the package returned the error from cleaning up the previous message. 

I believe the problem is that [flatWriteWrapper.Close](https://github.com/gorilla/websocket/blob/6f5d2139f4b96d5edfe6c1aac4fe87fed8d7f9fd/compression.go#L110) returns an [error when called more than once](https://github.com/gorilla/websocket/blob/6f5d2139f4b96d5edfe6c1aac4fe87fed8d7f9fd/compression.go#L110). That error should be ignored, but now all errors are ignored.  Perhaps that specific error should be ignored and all other errors should be returned.

**Summary:**

- Ignore error retuned from ` r.Close()` in [flatReadWrapper.Read](https://github.com/gorilla/websocket/blob/6f5d2139f4b96d5edfe6c1aac4fe87fed8d7f9fd/compression.go#L138C12-L138C22).

- Consider returning all errors from the call to c.writer.Close() except [errWriteClosed](https://github.com/gorilla/websocket/blob/6f5d2139f4b96d5edfe6c1aac4fe87fed8d7f9fd/conn.go#L180).

bcreane:
@pennystevens thank you for the thoughtful comments - I learned a bit more about the internals as a result of you leading me through the various interfaces and their behaviors. I elected to go with the "safe" approach which is to continue to ignore the errors returned via `c.writer.Close()`.

pennystevens:
More on c.writer.Close():  

The first writer error on the connection is [saved in Conn.writeErr](https://github.com/gorilla/websocket/blob/6f5d2139f4b96d5edfe6c1aac4fe87fed8d7f9fd/conn.go#L367).  

If c.writer.Close() returned an error because write to the connection failed, then the error is saved in  Conn.writerErr.  

Conn.beginMessage [returns that saved error](https://github.com/gorilla/websocket/blob/6f5d2139f4b96d5edfe6c1aac4fe87fed8d7f9fd/conn.go#L503-L508).

It is therefore safe to ignore the error returned from c.writer.Close().

DavidJohnSmith (approved):
I also find this problem,thanks for your commit

AlexVulaj (approved):

PR #879, opened 2023-12-08
perf: reduce timer in write_control

rfyiamcool:

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Documentation Update
- [ ] Go Version Update
- [ ] Dependency Update

## Description

In most cases, we do not write concurrently to the same client, and the `gorilla` library limits the concurrency, when `gorilla` detect write data concurrently, throw panic ?  In the  addition, creating and stopping a timer is not a simple action. 😁

So, we can get the lock first and if we can't get it, we can create a timer. 😁

## Run verifications and test

- [x] `make test` is passing

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -454,13 +454,18 @@ func (c *Conn) WriteControl(messageType int, data []byte, deadline time.Time) er
 		}
 	}
 
-	timer := time.NewTimer(d)
 	select {
 	case <-c.mu:
-		timer.Stop()
-	case <-timer.C:
-		return errWriteTimeout
+	default:
+		timer := time.NewTimer(d)
+		select {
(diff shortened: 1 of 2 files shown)


ethanwater (commented):
I like this!

AlexVulaj (approved):
Thanks for this update!

AlexVulaj:
@rfyiamcool  Before merging - would you mind adding/updating tests for this change?

rfyiamcool:
> @rfyiamcool Before merging - would you mind adding/updating tests for this change?

done, the ut is ok ? 😁

AlexVulaj (approved):

PR #883, opened 2023-12-14
fix leaking goroutine in chat example

benjamonnguyen:

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [x] Documentation Update
- [ ] Go Version Update
- [ ] Dependency Update

## Description
Fix leaking goroutine in the chat example that prevents http server from releasing port.

## Added/updated tests?

- [ ] Yes
- [x] No, and this is why: documentation

## Run verifications and test

- [x] `make test` is passing

### DIFF
--- a/examples/chat/hub.go
+++ b/examples/chat/hub.go
@@ -4,6 +4,10 @@
 
 package main
 
+import (
+	"context"
+)
+
 // Hub maintains the set of active clients and broadcasts messages to the
 // clients.
 type Hub struct {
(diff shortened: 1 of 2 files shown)


GreenMarmot:
The `main()` function always calls [log.Fatal](https://pkg.go.dev/log#Fatal) because [http.ListenAndServe](https://pkg.go.dev/net/http#ListenAndServe) always returns a non-nil error.  The log.Fatal function calls [os.Exit](https://pkg.go.dev/os#Exit).  The os.Exit function terminates the program immediately; deferred functions are not run.  The deferred call to `cancel()` is never run.  The PR is a no-op. 

There is not a leak because because the hub goroutine should run for the duration of the program.

The program can be improved by removing the if statement from main.  Change:

	err := server.ListenAndServe()
	if err != nil {
		log.Fatal("ListenAndServe: ", err)
	}

to:

	err := server.ListenAndServe()
	log.Fatal("ListenAndServe: ", err)

benjamonnguyen:
@GreenMarmot Thank you, I completely misconstrued another bug of my own making..

PR #889, opened 2024-01-23
feat: format message type

rfyiamcool:
### summary

The `messageType` and `message` can be retrieved using the readMessage method, i want to convert mt int to string by  `FormatMessageType` ? 😁 

```go
var upgrader = websocket.Upgrader{} // use default options

func echo(w http.ResponseWriter, r *http.Request) {
	c, err := upgrader.Upgrade(w, r, nil)
	if err != nil {
		log.Print("upgrade:", err)
		return
	}
	defer c.Close()
	for {
		mt, message, err := c.ReadMessage()
		if err != nil {
			log.Println("read:", err)
			break
		}

		log.Printf("recv: %s, type: %s", message, websocket.FormatMessageType(mt))  // 😁

		err = c.WriteMessage(mt, message)
		if err != nil {
			log.Println("write:", err)
			break
		}
	}
}
```

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -1263,3 +1263,15 @@ func FormatCloseMessage(closeCode int, text string) []byte {
 	copy(buf[2:], text)
 	return buf
 }
+
+var messageTypes = map[int]string{
+	TextMessage:   "TextMessage",
+	BinaryMessage: "BinaryMessage",
+	CloseMessage:  "CloseMessage",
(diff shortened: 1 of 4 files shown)


AlexVulaj:
Nice addition @rfyiamcool , thanks! Looks good to me.

AlexVulaj (approved):

Issue #877, opened 2023-11-28
[BUG] 1.5.1: use of closed network connection, 1.5.0 is fine


bmhatfield:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

After upgrading from websocket 1.5.0 to 1.5.1, my service started regularly erroring with `read tcp 10.40.0.207:8080->10.40.0.1:37429: use of closed network connection`. Reverting back to 1.5.0 from 1.5.1 quieted these errors.

### Expected Behavior

I expected 1.5.0 -> 1.5.1 to be a seamless upgrade, continuing to not error about use of closed network connection.

### Steps To Reproduce

This may be hard to reproduce, as I am unsure if this was just unmasked in our code or if a new issue was introduced to gorilla/websocket in 1.5.1. I will update this issue if I am able to find a clean reproduce case. We have _quite a bit_ of supporting code so the bug here could be entirely in our stack, not in gorilla/websocket.

However, a direct upgrade and then revert with no other corresponding changes which results in errors and then those errors going away seems like a good threshold to at least start a conversation.

### Anything else?

I reviewed the existing issues and found:
- https://github.com/gorilla/websocket/issues/48
- https://github.com/gorilla/websocket/issues/439

However both these issues predate 1.5.0 and 1.5.1 by some margin.

pennystevens:
- Are these errors returned from a connection method call or are these message written to the application's stderr?
- Does the application enable compression?  If no, then this reduces the number of changes to examine when hunting for this problem.

simon28082:
I've had the same problem recently, I'll try to roll back to 1.15.0

simon28082:
My own problem, solved.

Hayden-Chang:
> My own problem, solved.

Hi simon, could you describle how did you trigger this error and how did you resolved it?

StevenACoffman:
@simon28082 I'm curious if you can summarize what caused and solved your problem, as I'm experiencing the same symptoms.

mesidex:
I'm experiencing this as well while performing concurrent writes to my external websocket.

Enviy:
I was also encountering this issue. Despite checking for the net.ErrClosed error with errors.Is(), it was not being correctly caught.
As strange as it sounds, it was resolved by calling `go get github.com/gorilla/websocket@latest`. This was strange because the package version in go.mod did not change. A real riddle me this scenario.

jaitaiwan:
[...]

PR #872, opened 2023-11-22
make tests parallel

ninedraft:

PR #894, opened 2024-02-07
Remove hideTempErr to allow downstream users to check for errors like net.ErrClosed

UnAfraid:

- [x] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Go Version Update
- [ ] Dependency Update

## Description
Since this change https://github.com/gorilla/websocket/pull/840/files#diff-4f427d2b022907c552328e63f137561f6de92396d7a6e8f6c2ea1bcf0db52654L190-R197 we can no longer determinate if the errors coming from ReadMessage() are net.ErrClosed for example
Hardcoding the error message is not great option because it may vary from OS to OS and system locale

- Related Issue # https://github.com/gorilla/websocket/issues/877
- Closes # https://github.com/gorilla/websocket/issues/877

## Added/updated tests?

- [X] Yes
      have not been included_

## Run verifications and test

- [X] `make verify` is passing
- [X] `make test` is passing

### DIFF
--- a/conn_test.go
+++ b/conn_test.go
@@ -823,7 +823,6 @@ func (r fakeNetClosedReader) Read([]byte) (int, error) {
 }
 
 func TestConnectionClosed(t *testing.T) {
-	t.Parallel()
 	var b1, b2 bytes.Buffer
 
 	client := newTestConn(fakeNetClosedReader{}, &b1, false)
--- a/conn_test.go
+++ b/conn_test.go
@@ -814,3 +814,26 @@ func TestFormatMessageType(t *testing.T) {
 		t.Error("failed to format message type")
 	}
 }
+
+type fakeNetClosedReader struct {
+}
+
+func (r fakeNetClosedReader) Read([]byte) (int, error) {
+	return 0, net.ErrClosed
+}
+
+func TestConnectionClosed(t *testing.T) {
+	t.Parallel()
+	var b1, b2 bytes.Buffer
+
+	client := newTestConn(fakeNetClosedReader{}, &b1, false)
+	server := newTestConn(fakeNetClosedReader{}, &b2, true)
+
+	if _, _, err := server.NextReader(); !errors.Is(err, net.ErrClosed) {
+		t.Fatalf("server expects a net.ErrClosed error, %v returned", err)
+	}
+
+	if _, _, err := client.NextReader(); !errors.Is(err, net.ErrClosed) {
+		t.Fatalf("client expects a net.ErrClosed error, %v returned", err)
+	}
+}
--- a/conn.go
+++ b/conn.go
@@ -192,13 +192,6 @@ func newMaskKey() [4]byte {
 	return k
 }
 
-func hideTempErr(err error) error {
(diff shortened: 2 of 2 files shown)


GreenMarmot:
Thank you! The maintainer's decision to hide all errors from the application was mystifying and is one of the reasons we have not upgraded past 1.5.0.

StevenACoffman:
@AlexVulaj This would fix several issues with the 1.5.1 release. Please take a look. Thanks!

conn_test.go
```diff
@@ -797,3 +797,26 @@ func TestFailedConnectionReadPanic(t *testing.T) {
...
+
+type fakeNetClosedReader struct {
+}
+
+func (r fakeNetClosedReader) Read([]byte) (int, error) {
+	return 0, net.ErrClosed
+}
+
+func TestConnectionClosed(t *testing.T) {
+	t.Parallel()
```
jaitaiwan, line 809: I'm not sure if we really need this test to be run in Parallel?

jaitaiwan (approved):
LGTM, one small nit. I am a little cautious about this change because I haven't looked at the context of c.readErr. If other Members are happy then I don't see any structural issues with this.

GreenMarmot:
Prior to commit 666c197fc9157896b57515c3a3326c3f8c8319fe, hideTempError wrapped errors with e.Temporary() == true.  All other errors were passed through as is.  The PR that added hideTempError says "Do not return net.Error with Temporary() == true".   This was probably done because errors on the connection are not recoverable.
 
The net.Error Temporary() method in the standard library was recently deprecated because the meaning of the method was not well-defined. 

This PR brings the package closer to 1.50 by not wrapping any errors.

Regarding the parallel test:  #872 sprinkled t.Parallel() through almost all tests, including [tests where it makes no sense](https://github.com/gorilla/websocket/blob/d08ee1ad9b8a1bad1b388165f85f7794a604b8e1/example_test.go#L45).  For better or worse, the call to t.Parallel() goes with the flow.

conn_test.go
UnAfraid: Removed t.Parallel

AlexVulaj: I could go either way but I'm in favor of leaving the parallel test for consistency since https://github.com/gorilla/websocket/pull/872 was recently merged.

AlexVulaj:
I'm happy to merge this @jaitaiwan , feel free to make a final call on the "t.Parallel" comment but I don't have any strong feelings.

UnAfraid:
Please let me know if t.Parallel() should be used and i can bring it back, i don't mind either way.

conn_test.go
jaitaiwan: Lets leave it as is and not require more changes.

jaitaiwan (approved):
LGTM

StevenACoffman:
Test and lint failures are unrelated to this change.

jaitaiwan:
Agreed. Thanks for the contribution!

PR #896, opened 2024-02-13
Remove `done` channel in client-server echo example

CicadaCinema:

- [x] Refactor
- [ ] Feature
- [ ] Bug Fix
- [x] Optimization
- [x] Documentation Update
- [ ] Go Version Update
- [ ] Dependency Update

## Description

I don't believe anything is actually sent to the `done` channel, in this example there are only statements reading from it. Furthermore, the deferred `close(done)` statement will never execute because deferred statements in other goroutines are not run when main terminates. So no messages will ever be sent to this channel and it will never be closed, so it is safe to remove all the references to `done`.

I took a look at https://github.com/gorilla/websocket/commit/844dd6d40e1a9215ef4c8a204bfc839fcf5dd5dd , where `done` was introduced, and the situation was similar then, so I believe this is an oversight in that commit.

None

## Added/updated tests?

- [ ] Yes
- [x] No, and this is why: I only slightly modified some code in an example application, and the functionality remains unaffected so there is nothing more to test

## Run verifications and test

- [ ] `make test` is passing

I didn't run these ^ .

### DIFF
--- a/examples/echo/client.go
+++ b/examples/echo/client.go
@@ -36,10 +36,7 @@ func main() {
 	}
 	defer c.Close()
 
-	done := make(chan struct{})
-
 	go func() {
-		defer close(done)
 		for {
 			_, message, err := c.ReadMessage()
 			if err != nil {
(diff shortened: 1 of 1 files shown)


CicadaCinema:
Nevermind, I realised that the goroutine will actually terminate due to an error coming from ReadMessage().

PR #900, opened 2024-02-26
Fix: Calling the Close method from multiple goroutines concurrently may result in race conditions.

yc90s:
Calling the Close method from multiple goroutines concurrently may result in race conditions, using sync.Once instead

### DIFF
--- a/examples/chat/client.go
+++ b/examples/chat/client.go
@@ -8,6 +8,7 @@ import (
 	"bytes"
 	"log"
 	"net/http"
+	"sync"
 	"time"
 
 	"github.com/gorilla/websocket"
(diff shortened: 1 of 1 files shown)


GreenMarmot:
This change is unnecessary.  

The [package documentation](https://pkg.go.dev/github.com/gorilla/websocket#hdr-Concurrency) says that the Close() method can be called concurrently.

The [close method implementation](https://github.com/gorilla/websocket/blob/695e9095ce8736ac99c83939ca6b0fe93768f680/conn.go#L343-L345) is a single call to Close() on the network connection.  Network connections support concurrent calls to Close().

PR #902, opened 2024-03-06
examples/echo changed variables for clarity and removed nonexistent method

datadius:

- [ ] Refactor
- [ ] Feature
- [X] Bug Fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Go Version Update
- [ ] Dependency Update

## Description
In the example/echo folder, an error has been introduced where a nonexistent function was used. I removed the error and renamed the variables for better clarity.

The variable was messageType that comes from conn.ReadMessage(). messageType is an int with value websocket.BinaryMessage or websocket.TextMessage. Before it was marked as _, but I believe the example would be improved if we used it in a simple way to show that it can be used as an indicator for the type of message.

I couldn't find related issues.

## Added/updated tests?

- [ ] Yes
- [X] No, and this is why: I just modified the example for the new users.

## Run verifications and test

- [X] `make verify` is passing
- [X] `make test` is passing

I ran them anyway, even though I didn't change any production code

### DIFF
--- a/examples/echo/client.go
+++ b/examples/echo/client.go
@@ -41,12 +41,16 @@ func main() {
 	go func() {
 		defer close(done)
 		for {
-			mt, message, err := c.ReadMessage()
+			messageType, message, err := c.ReadMessage()
 			if err != nil {
 				log.Println("read:", err)
 				return
 			}
-			log.Printf("recv: %s, type: %s", message, websocket.FormatMessageType(mt))
+			if messageType == websocket.TextMessage {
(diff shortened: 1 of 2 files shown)


GreenMarmot:
>  an error has been introduced where a nonexistent function was used

The FormatMessage function [exists](https://github.com/gorilla/websocket/blob/b2a86a174473e870c9fcc3eaba8957689090d2cc/conn.go#L1268).

PR #912, opened 2024-04-25
feat: add method to acknowledge expected/temporary read errors

### LINKED ISSUE #476: panic: repeated read on failed websocket connection
type WsConn struct {
	*websocket.Conn
	proxy           string
	host            string
	path            string
	closed          bool
            ....................
}

func (ws *WsConn) StartToReceiveMessage() {
	go func() {
		for {
			t, msg, err := ws.ReadMessage()  // this line caused the panic
			if ws.closed {
				glog.V(7).Infof("websocket closed.....")
				return
			}
			ws.fireMessage(err, t, msg)
		}
	}()
}

I create the connection to the server and subsribe some api.       
run after few seconds.....

panic: repeated read on failed websocket connection

goroutine 69 [running]:
github.com/gorilla/websocket.(*Conn).NextReader(0xc00017a160, 0xc000092008, 0xc0003f1fa8, 0x2, 0x2, 0xb)
	D:/GoWork/src/github.com/gorilla/websocket/conn.go:968 +0x35d
github.com/gorilla/websocket.(*Conn).ReadMessage(0xc00017a160, 0x2, 0x2, 0xb, 0x0, 0x0, 0xc0002402c0)
	D:/GoWork/src/github.com/gorilla/websocket/conn.go:1029 +0x36

Process finished with exit code 2

dpeckett:

- [ ] Refactor
- [x] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Go Version Update
- [ ] Dependency Update

## Description

Add a method for explicitly acknowledging read errors are temporary. This is useful when read timeouts are expected, such as in a non blocking loop with context cancellation etc.

- Closes #476 

## Added/updated tests?

- [ ] Yes
- [x] No, and this is why: negligible code changes

## Run verifications and test

- [x] `make test` is passing

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -1007,8 +1007,9 @@ func (c *Conn) handleProtocolError(message string) error {
 //
 // Applications must break out of the application's read loop when this method
 // returns a non-nil error value. Errors returned from this method are
-// permanent. Once this method returns a non-nil error, all subsequent calls to
-// this method return the same error.
+// permanent, unless explicitly cleared using ClearReadError. Once this method
+// returns a non-nil error, all subsequent calls to this method return the
+// same error.
 func (c *Conn) NextReader() (messageType int, r io.Reader, err error) {
 	// Close previous reader, only relevant for decompression.
 	if c.reader != nil {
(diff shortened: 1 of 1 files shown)


GreenMarmot:
What is the scenario where this feature can be used reliably?

Timeout is not one of those scenarios.  Consider the case where the read deadline is exceeded when [reading the frame length](https://github.com/gorilla/websocket/blob/09a6bab4667fc302c7f23ad770ce48be6fc2d15f/conn.go#L893-L912).  The [first two bytes](https://github.com/gorilla/websocket/blob/09a6bab4667fc302c7f23ad770ce48be6fc2d15f/conn.go#L819) of the frame are discarded and all future reads on the network connection will be out of sync by two bytes.  

Rather than adding this band-aid, `advanceFrame`  should be modified to [peek](https://pkg.go.dev/bufio#Reader.Peek) through the frame header and only [consume](https://pkg.go.dev/bufio#Reader.Discard) the bytes on success.  With that change, `NextReader` can be modified to make the deadline exceeded error temporary.

Regarding context cancelation: [Close](https://pkg.go.dev/github.com/gorilla/websocket#Conn.Close) the connection when the context is canceled.  Any active calls to NextReader or ReadMessage will immediately return with an error.

PR #916, opened 2024-05-09
feat: register https proxy

### LINKED ISSUE #794: Is https not supported?
It looks like gorilla websocket intends to support https, e.g logic [here](https://github.com/gorilla/websocket/blob/9111bb834a68b893cebbbaed5060bdbc1d9ab7d2/client.go#L259). 

However, https is not registered during init [here](https://github.com/gorilla/websocket/blob/9111bb834a68b893cebbbaed5060bdbc1d9ab7d2/proxy.go#L24). 

This caused https to be considered as an unknown proxy when we use https proxy, error thrown from [here](https://github.com/gorilla/websocket/blob/9111bb834a68b893cebbbaed5060bdbc1d9ab7d2/x_net_proxy.go#L237). 

Is this a bug? Shouldn't the `proxy_RegisterDialerType` be initialized with `https`?

jackstar12:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Go Version Update
- [ ] Dependency Update

## Description

This PR registers a `https` proxy dialer type. I've tested it manually, but I'm not sure if / how to approach tests.

- Closes #794

## Added/updated tests?

- [ ] Yes
      have not been included_

## Run verifications and test

- [ ] `make verify` is passing (not passing on main either)
- [x] `make test` is passing

### DIFF
--- a/proxy.go
+++ b/proxy.go
@@ -23,9 +23,11 @@ func (fn netDialerFunc) Dial(network, addr string) (net.Conn, error) {
 }
 
 func init() {
-	proxy.RegisterDialerType("http", func(proxyURL *url.URL, forwardDialer proxy.Dialer) (proxy.Dialer, error) {
+	generate := func(proxyURL *url.URL, forwardDialer proxy.Dialer) (proxy.Dialer, error) {
 		return &httpProxyDialer{proxyURL: proxyURL, forwardDial: forwardDialer.Dial}, nil
-	})
+	}
+	proxy.RegisterDialerType("http", generate)
+	proxy.RegisterDialerType("https", generate)
 }
 
 type httpProxyDialer struct {


canelohill:
HTTPS proxies are not supported by the websocket package.  The current code returns an error when the application [specifies](https://pkg.go.dev/github.com/gorilla/websocket#Dialer) an HTTPS proxy.

The code in this commit silently downgrades to HTTP when the application specifies an HTTPS proxy.  That's dangerous from a security standpoint.

Reject this PR.

jaitaiwan:
Agreed thanks @canelohill and appreciate the effort @jackstar12.

PR #917, opened 2024-05-19
command line chat example

IyadElwy:

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [x] Documentation Update
- [ ] Go Version Update
- [ ] Dependency Update

## Description

Example of a quick chat app which, unlike the current chat app does not require any HTML or browser and be run directly using two open terminal sessions

- Related Issue #
- Closes #

## Added/updated tests?

- [ ] Yes
- [ ] No, and this is why: this is just an example

## Run verifications and test

- [ ] `make test` is passing

### DIFF
--- a/examples/cmd-chat/client/main.go
+++ b/examples/cmd-chat/client/main.go
@@ -1,0 +1,81 @@
+package main
+
+import (
+	"bufio"
+	"fmt"
+	"math/rand"
+	"net/http"
+	"os"
+	"os/signal"
+
+	"github.com/gorilla/websocket"
+)
+
+func main() {
+	connUserId := rand.Intn(100)
+	header := http.Header{}
+	header.Add("user-id", fmt.Sprint(connUserId))
+
+	dialer := websocket.Dialer{}
+	conn, _, err := dialer.Dial("ws://localhost:4000", header)
+	if err != nil {
+		err := conn.WriteMessage(websocket.CloseMessage,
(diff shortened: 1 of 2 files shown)


nanoblackstone:
The server application has a data race.  To observe the problem, run the server with the race detector and connect with two or more clients.

The handling of close messages is clunky.  The server disconnects a client when there's an error writing a message to another client.  Fix both of these issues with the following:

        // Delete call to and code for alterCloseHandlerToRemoveConnections 
	go func() {
		defer func() {
			conn.Close()
			// next line is copied from original, but is involved with data race.
			delete(connections, connUserId)
		}()

		for {
			_, p, err := conn.ReadMessage()
			if err != nil {
				if websocket.IsCloseError(err, websocket.CloseNormalClosure) {
					broadcastMessage(fmt.Sprintf("User %s left the chat\n", connUserId), connUserId)
				} else {
					broadcastMessage(fmt.Sprintf("User %s disconnected unexpectedly\n", connUserId), connUserId)
				}
				return
			}
			broadcastMessage(fmt.Sprintf("%s: %s", connUserId, p), connUserId)
		}
	}()

Does this example cover a websocket package feature not covered by another example?

AlexVulaj:
Hey @IyadElwy , thanks for putting this together. I appreciate the time and effort you put into this, but I don't think this particular example adds anything beyond the existing "chat" example that already exists within the project. I might recommend an external blog post or something similar to showcase your hard work.

PR #932, opened 2024-06-18
Handle errcheck warnings

canelohill:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Go Version Update
- [ ] Dependency Update

## Description

The package ignored errors from net.Conn Set*Deadline in a few places. Update the package to return these errors to the caller.

Ignore all other errors reported by errcheck. These errors are safe to ignore because
- The function is making a best effort to cleanup while handling another error.
- The function call is guaranteed to succeed.
- The error is ignored in a test.

- Related Issue #
- Closes #

## Added/updated tests?

- [ ] Yes
- [x] No, and this is why: covered by existing tests

## Run verifications and test

- [ ] `make test` is passing

### DIFF
--- a/client.go
+++ b/client.go
@@ -305,9 +305,15 @@ func (d *Dialer) DialContext(ctx context.Context, urlStr string, requestHeader h
 		})
 	}
 
+	// Close the network connection when returning an error. The variable
+	// netConn is set to nil before the success return at the end of the
+	// function.
 	defer func() {
 		if netConn != nil {
-			netConn.Close()
+			// It's safe to ignore the error from Close() because this code is
+			// only executed when returning a more important error to the
+			// application.
+			_ = netConn.Close()
 		}
 	}()
 
@@ -398,8 +404,14 @@ func (d *Dialer) DialContext(ctx context.Context, urlStr string, requestHeader h
 	resp.Body = io.NopCloser(bytes.NewReader([]byte{}))
 	conn.subprotocol = resp.Header.Get("Sec-Websocket-Protocol")
 
-	netConn.SetDeadline(time.Time{})
-	netConn = nil // to avoid close in defer.
+	if err := netConn.SetDeadline(time.Time{}); err != nil {
+		return nil, resp, err
+	}
+
+	// Success! Set netConn to nil to stop the deferred function above from
+	// closing the network connection.
+	netConn = nil
+
 	return conn, resp, nil
 }
 
--- a/conn.go
+++ b/conn.go
@@ -370,7 +370,9 @@ func (c *Conn) read(n int) ([]byte, error) {
 	if err == io.EOF {
 		err = errUnexpectedEOF
 	}
-	c.br.Discard(len(p))
+	// Discard is guaranteed to succeed because the number of bytes to discard
+	// is less than or equal to the number of bytes buffered.
+	_, _ = c.br.Discard(len(p))
 	return p, err
 }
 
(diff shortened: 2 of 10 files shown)


conn.go
```diff
@@ -934,7 +938,7 @@ func (c *Conn) advanceFrame() (int, error) {
 	var payload []byte
 	if c.readRemaining > 0 {
 		payload, err = c.read(int(c.readRemaining))
-		c.setReadRemaining(0)
+		_ = c.setReadRemaining(0) // will not overflow
```
jaitaiwan, line 941: Once again, I don't trust this comment, a why would be helpful

conn.go
```diff
@@ -1054,7 +1059,7 @@ func (r *messageReader) Read(b []byte) (int, error) {
 			}
 			rem := c.readRemaining
 			rem -= int64(n)
-			c.setReadRemaining(rem)
+			_ = c.setReadRemaining(rem) // will not overflow
```
jaitaiwan, line 1062: I don't trust this comment even more.

server.go
```diff
@@ -182,8 +182,16 @@ func (u *Upgrader) Upgrade(w http.ResponseWriter, r *http.Request, responseHeade
 		return u.returnError(w, r, http.StatusInternalServerError, err.Error())
 	}
 
+	defer func() {
+		if netConn != nil {
+			// It's safe to ignore the error from Close() because this code is
+			// only executed when returning a more important to the
```
jaitaiwan, line 188: I think the word error is missing from this sentence

server.go
```diff
@@ -247,20 +255,30 @@ func (u *Upgrader) Upgrade(w http.ResponseWriter, r *http.Request, responseHeade
...
 	if u.HandshakeTimeout > 0 {
-		netConn.SetWriteDeadline(time.Time{})
+		if err := netConn.SetWriteDeadline(time.Time{}); err != nil {
+			return nil, err
+		}
 	}
 
+	// Set netConn to nil to avoid call to netConn.Close() in
+	// deferred function call.
+	netConn = nil
```
jaitaiwan, line 280: I feel like it should be more explicit that we're setting netConn to nil because we're expecting the closure due to the instant SetDeadline. If that's not the correct understanding then it's even more confusing...

jaitaiwan (changes requested):
I have a few pointers on this that I think need solving before merge.

client.go
```diff
@@ -421,8 +421,14 @@ func (d *Dialer) DialContext(ctx context.Context, urlStr string, requestHeader h
 	resp.Body = ioutil.NopCloser(bytes.NewReader([]byte{}))
 	conn.subprotocol = resp.Header.Get("Sec-Websocket-Protocol")
 
-	netConn.SetDeadline(time.Time{})
-	netConn = nil // to avoid close in defer.
+	if err := netConn.SetDeadline(time.Time{}); err != nil {
+		return nil, nil, err
+	}
```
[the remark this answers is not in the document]
canelohill: The response will not aid debugging the timeout error, but it does not hurt to include it in the result.

server.go
canelohill: I updated Upgrader.Upgrade to match the pattern in Dialer.DialContext:  A deferred function ensures that the underlying network connection is closed on all error return paths.  I improved the comments in both functions.

jaitaiwan (dismissed):

jaitaiwan:
@canelohill I had to manually resolve a conflict. Can you check and make sure that fits what you think should happen?

jaitaiwan (approved):

jaitaiwan (approved):

canelohill:
The PR looks good after merges.

jaitaiwan:
@canelohill  can’t seem to merge it says there’s conflicts

jaitaiwan:
Cheers thanks for that. I was surprised cause usually I get the opportunity to resolve conflicts manually but this time there was no option.

shahshahii:
hi🫡

PR #938, opened 2024-06-19
Unbundle x/net/proxy and update to recent version

canelohill:

- [x] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Go Version Update
- [x] Dependency Update

## Description

Import golang.org/x/net/proxy instead of using the bundle in x_net_proxy.go. There's no need to avoid the dependency on golang.org/x/net/proxy now that Go's module system is in widespread use.

Change Dialer.DialContext to pass contexts as an argument to the dial function instead of tunneling the context through closures. Tunneling is no longer needed because the proxy package supports contexts. The version of the proxy package in the bundle predates contexts!

Simplify the code for calculating the base dial function.

Prevent the HTTP proxy dialer from leaking out of the websocket package by selecting the HTTP proxy dialer directly in the websocket package. Previously, the HTTP dialer was registered with the proxy package.

## Added/updated tests?

- [ ] Yes
- [x] No, and this is why: covered by existing tests

## Run verifications and test

- [ ] `make test` is passing


AlexVulaj:
Very in favor of this - approved. Thanks for the PR!

AlexVulaj (approved):

PR #942, opened 2024-06-23
Improve bufio handling in Upgrader.Upgrade

### LINKED ISSUE #642: Upgrader.Upgrade not writing error response to http.ResponseWriter for invalid request with payload data
**Describe the bug**
I'm running a websocket server that uses Upgrader.Upgrade() to handle handshake requests for switching protocols from http to websocket connection.
Upgrader.Upgrade() seems to validate all websocket headers and writes error code with response to http.ResponseWriter if any required header is missing or invalid.
But if the client sends a request with unnecessary payload data, Upgrader.Upgrade doesn't send an error response to the client.
It however returns an error to the service saying `websocket: client sent data before handshake is complete`. And the client shuts without a reply from server

On receiving this error in the server, I tried to send an error response to the client explicitly. But I can't, as the http connection is deemed 'hijacked'.

**Problem**: How should the server handle request with payload data? Do I need to validate that an incoming handshake request should have no payload, before using Upgrader.Upgrade to switch protocols?
If not, how to I send an error code to the client.
…

**Versions**
Go version: 1.11.5
package version: b65e62901fc1c0d968042419e74789f6af455eb9

…

**Steps to Reproduce**
Use Upgrader.Upgrade() to upgrade an http GET request. Send payload data in the client request. 

…

**Expected behavior**
A proper error code and response must be sent to the client for a bad handshake request if it contains an unnecessary payload data.

…

**Code Snippets**

	h, ok := w.(http.Hijacker)
	if !ok {
		return u.returnError(w, r, http.StatusInternalServerError, "websocket: response does not implement http.Hijacker")
	}
	var brw *bufio.ReadWriter
	netConn, brw, err := h.Hijack()
	if err != nil {
		return u.returnError(w, r, http.StatusInternalServerError, err.Error())
	}

	if brw.Reader.Buffered() > 0 {
		netConn.Close()
		return nil, errors.New("websocket: client sent data before handshake is complete")
	}
Originating here: https://github.com/gorilla/websocket/blob/c3dd95aea9779669bb3daafbd84ee0530c8ce1c1/server.go#L185

…

canelohill:

- [x] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update
- [x] Go Version Update
- [ ] Dependency Update

## Description

Use Reader.Size() (add in Go 1.10) to get the bufio.Reader's size instead of examining the return value from Reader.Peek.

Use Writer.AvailableBuffer() (added in Go 1.18) to get the bufio.Writer's buffer instead of observing the buffer in the underlying writer.

Allow client to send data before the handshake is complete. Previously, Upgrader.Upgrade rudely closed the connection.

- Closes #642

## Added/updated tests?

- [x] Yes
      have not been included_

## Run verifications and test

- [ ] `make test` is passing

### DIFF
--- a/client_server_test.go
+++ b/client_server_test.go
@@ -5,6 +5,7 @@
 package websocket
 
 import (
+	"bufio"
 	"bytes"
 	"context"
 	"crypto/tls"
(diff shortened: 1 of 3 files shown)


jaitaiwan (approved):

canelohill:
@naesheim The PR allows a client to send messages before the handshake is complete.  Otherwise, the PR is cleanup. 

Because the WebSocket RFC does not allow a client to send messages before the handshake is complete, this PR is only essential when working with a broken client.

PR #950, opened 2024-07-18
Support TLS connection to HTTP Proxy

### LINKED ISSUE #739: [feature] Support for proxying websocket through https proxy
**Is your feature request related to a problem? Please describe.**

I need to use a proxy server that itself requires TLS (proxy URL has https scheme).

**Describe the solution you'd like**

I should be able to use a proxy server (eg, via ProxyFromEnvironment) that has an https scheme and uses TLS. Currently, `proxy_RegisterDialerType` only registers the 'http' scheme.
This should also be able to use my client `TLSClientConfig` which could be necessary to communicate with my proxy.

**Describe alternatives you've considered**
I do not see any other alternatives. The workaround suggested in comments below on 12/3/21 is impractical, especially when using `ProxyFromEnvironment` and applying this to other libraries. I should not have to mangle my proxy scheme to get this to work.

### LINKED ISSUE #794: Is https not supported?

134130:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Documentation Update
- [ ] Go Version Update
- [ ] Dependency Update

## Description
- Support TLS Connection between websocket client and http proxy.

cc. @adrianosela

- Closes #739
- Closes #794

## Added/updated tests?

- [x] Yes
      have not been included_

## Run verifications and test

- [x] `make test` is passing

### DIFF
--- a/client.go
+++ b/client.go
@@ -368,7 +368,7 @@ func (d *Dialer) DialContext(ctx context.Context, urlStr string, requestHeader h
 				if proto != "http/1.1" {
 					return nil, nil, fmt.Errorf(
 						"websocket: protocol %q was given but is not supported;"+
-							"sharing tlsServerName.Config with net/http Transport can cause this error: %w",
+							"sharing tls.Config with net/http Transport can cause this error: %w",
 						proto, err,
 					)
 				}
--- a/client.go
+++ b/client.go
@@ -246,19 +246,7 @@ func (d *Dialer) DialContext(ctx context.Context, urlStr string, requestHeader h
 		defer cancel()
 	}
 
-	var netDial netDialerFunc
-	switch {
-	case u.Scheme == "https" && d.NetDialTLSContext != nil:
-		netDial = d.NetDialTLSContext
-	case d.NetDialContext != nil:
-		netDial = d.NetDialContext
-	case d.NetDial != nil:
-		netDial = func(ctx context.Context, net, addr string) (net.Conn, error) {
-			return d.NetDial(net, addr)
-		}
-	default:
-		netDial = (&net.Dialer{}).DialContext
-	}
+	netDial := newNetDialerFunc(u.Scheme, d.NetDial, d.NetDialContext, d.NetDialTLSContext)
 
 	// If needed, wrap the dial function to connect through a proxy.
 	if d.Proxy != nil {
--- a/client.go
+++ b/client.go
@@ -17,6 +17,8 @@ import (
 	"net/url"
 	"strings"
 	"time"
+
+	"golang.org/x/net/proxy"
 )
 
 // ErrBadHandshake is returned when the server response to opening handshake is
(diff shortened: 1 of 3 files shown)


client.go
```diff
@@ -275,20 +298,6 @@ func (d *Dialer) DialContext(ctx context.Context, urlStr string, requestHeader h
...
-		if err != nil {
-			return nil, nil, err
-		}
-		if proxyURL != nil {
-			netDial, err = proxyFromURL(proxyURL, netDial)
-			if err != nil {
-				return nil, nil, err
-			}
-		}
-	}
```
134130, line 290: Also need to set a deadline for the dialing to proxy, so moved the line up.

client_server_test.go
```diff
@@ -85,6 +85,51 @@ func newTLSServer(t *testing.T) *cstServer {
...
+}
+
+func newProxyServer() *httptest.Server {
+	return httptest.NewServer(&cstProxyServer{})
+}
+
+func newTLSProxyServer() *httptest.Server {
+	return httptest.NewTLSServer(&cstProxyServer{})
+}
+
```
134130, line 132: Previously, only had a temporary proxy implementation, and couldn't test the TLS proxy, so implemented it.

client_server_test.go
```diff
@@ -85,6 +85,51 @@ func newTLSServer(t *testing.T) *cstServer {
...
+	if err != nil {
+		_, _ = fmt.Fprintf(conn, "HTTP/1.1 502 Bad Gateway\r\n\r\n")
+		return
+	}
+	defer upstream.Close()
+
+	_, _ = fmt.Fprintf(conn, "HTTP/1.1 200 Connection established\r\n\r\n")
+
+	wg := sync.WaitGroup{}
+	wg.Add(2)
```
adrianosela, line 113: Suggestion:  I think you probably want to exit as soon as one io.Copy hits EOF...

Here's two ways I think you can achieve this:

```
	var wg sync.WaitGroup
	defer wg.Wait()

	wg.Add(1)
	go func() {
		defer wg.Done()

		// abort blocked reads from upstream when done reading from conn
		defer upstream.SetDeadline(time.Now().Add(-1 * time.Hour))

		_, _ = io.Copy(upstream, conn)
	}()

	wg.Add(1)
	go func() {
		defer wg.Done()

		// abort blocked reads from conn when done reading from upstream
		defer conn.SetDeadline(time.Now().Add(-1 * time.Hour))

		_, _ = io.Copy(conn, upstream)
	}()
```

OR more simple - block on an unbuffered channel:

```
        done := make(chan struct{})

	go func() {
		_, _ = io.Copy(upstream, conn)
                done <- struct{}{}
	}()

	go func() {
		_, _ = io.Copy(conn, upstream)
               done <- struct{}{}
	}()

       <- done
```

adrianosela (approved):
Looks great :D

Thanks so much for hopping on this so quickly. Im wondering what kind of testing was done (aside from unit tests).

client_server_test.go
134130: Yes, I prefer the below. I used `wg` because It's just a test helper function.

I'll change to the below one.

134130:
Closing since it will be implemented in #978

PR #952, opened 2024-08-18
Use embeds for home files in chat examples

eldondev:
Rather than serving up the html files directly
from whatever directory the program is run in,
embed the files in the binary.

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [X] Documentation Update
- [ ] Go Version Update
- [ ] Dependency Update

## Added/updated tests?

- [X] No, and this is why: Just example code

## Run verifications and test 

- [X] I don't see any makefiles?
- [ ] `make test` is passing

### DIFF
--- a/examples/chat/main.go
+++ b/examples/chat/main.go
@@ -5,12 +5,18 @@
 package main
 
 import (
+	"bytes"
+	_ "embed"
 	"flag"
 	"log"
 	"net/http"
+	"time"
 )
 
 var addr = flag.String("addr", ":8080", "http service address")
+
+//go:embed home.html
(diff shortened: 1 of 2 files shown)


WofWca:
IDK IDK. It's pretty common for web servers to serve files instead of embedding them in the binary. This IMO complicates the example without a practical benefit.

hulkingshtick:
Benefits of this change are:
- the `cd` step for running the example can be deleted. 
- the example can be `go install`ed.  
- the example will follow current best practices for accessing static assets.

PR #967, opened 2024-11-28
Expose tls connection state in http response

004helix:
## What type of PR is this?

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Go Version Update
- [ ] Dependency Update

## Description

Expose TLS field in http.Response struct if tls handshake done in DialContext
https://pkg.go.dev/net/http@go1.20#Response

## Added/updated tests?

- [x] Yes
      have not been included_

## Run verifications and test

- [ ] `make test` is passing

### DIFF
--- a/client.go
+++ b/client.go
@@ -317,6 +317,7 @@ func (d *Dialer) DialContext(ctx context.Context, urlStr string, requestHeader h
 		}
 	}()
 
+	var tlsState *tls.ConnectionState
 	if u.Scheme == "https" && d.NetDialTLSContext == nil {
 		// If NetDialTLSContext is set, assume that the TLS handshake has already been done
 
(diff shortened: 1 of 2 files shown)


jack-flippen:
See also #1029.  Pick one.

client.go
```diff
@@ -330,13 +331,18 @@ func (d *Dialer) DialContext(ctx context.Context, urlStr string, requestHeader h
...
-			trace.TLSHandshakeDone(tlsConn.ConnectionState(), err)
-		}
 
-		if err != nil {
+		if err := doHandshake(ctx, tlsConn, cfg); err != nil {
+			if trace != nil && trace.TLSHandshakeDone != nil {
+				trace.TLSHandshakeDone(tls.ConnectionState{}, err)
+			}
 			return nil, nil, err
+		} else {
```
jack-flippen, line 340: Move the happy path out of the else block.  The if block always returns, so no need to put the code under else.

PR #978, opened 2025-02-27
Implements missing HTTPS proxy functionality

seans3:

- [X] Feature

## Description

* Implements missing HTTPS Proxy functionality.
* Adds client/proxy/server unit tests.

- Closes https://github.com/gorilla/websocket/issues/739
- Closes https://github.com/gorilla/websocket/issues/794

## Added/updated tests?

- [X] Yes

## Run verifications and test

- Previous Unit Test Coverage: 83.6%
- Current Unit Test Coverage: 84.2%

```
$ go test -race -cover
PASS
coverage: 84.2% of statements
ok  	github.com/gorilla/websocket	4.103s
```

### DIFF
--- a/client.go
+++ b/client.go
@@ -51,18 +51,34 @@ func NewClient(netConn net.Conn, u *url.URL, requestHeader http.Header, readBufS
 //
 // It is safe to call Dialer's methods concurrently.
 type Dialer struct {
+	// The following custom dial functions can be set to establish
+	// connections to either the backend server or the proxy (if it
+	// exists). The scheme of the dialed entity (either backend or
+	// proxy) determines which custom dial function is selected:
+	// either NetDialTLSContext for HTTPS or NetDialContext/NetDial
+	// for HTTP. Since the "Proxy" function can determine the scheme
+	// dynamically, it can make sense to set multiple custom dial
+	// functions simultaneously.
+	//
 	// NetDial specifies the dial function for creating TCP connections. If
 	// NetDial is nil, net.Dialer DialContext is used.
+	// If "Proxy" field is also set, this function dials the proxy--not
+	// the backend server.
 	NetDial func(network, addr string) (net.Conn, error)
 
 	// NetDialContext specifies the dial function for creating TCP connections. If
 	// NetDialContext is nil, NetDial is used.
+	// If "Proxy" field is also set, this function dials the proxy--not
+	// the backend server.
 	NetDialContext func(ctx context.Context, network, addr string) (net.Conn, error)
 
 	// NetDialTLSContext specifies the dial function for creating TLS/TCP connections. If
 	// NetDialTLSContext is nil, NetDialContext is used.
 	// If NetDialTLSContext is set, Dial assumes the TLS handshake is done there and
 	// TLSClientConfig is ignored.
+	// If "Proxy" field is also set, this function dials the proxy (and performs
+	// the TLS handshake with the proxy, ignoring TLSClientConfig). In this TLS proxy
+	// dialing case the TLSClientConfig could still be necessary for TLS to the backend server.
 	NetDialTLSContext func(ctx context.Context, network, addr string) (net.Conn, error)
 
 	// Proxy specifies a function to return a proxy for a given
@@ -73,7 +89,7 @@ type Dialer struct {
 
 	// TLSClientConfig specifies the TLS configuration to use with tls.Client.
 	// If nil, the default configuration is used.
-	// If either NetDialTLS or NetDialTLSContext are set, Dial assumes the TLS handshake
+	// If NetDialTLSContext is set, Dial assumes the TLS handshake
 	// is done there and TLSClientConfig is ignored.
 	TLSClientConfig *tls.Config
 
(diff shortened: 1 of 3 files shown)


adrianosela (approved):
Thanks for this @seans3 🚀

aojea:
IIUIC the same certificate is used for both the proxy and the backend, don't we need to handle the case when both are different? what happens if the proxy requires a different certificate than the backend?

client.go
```diff
@@ -415,6 +414,53 @@ func (d *Dialer) DialContext(ctx context.Context, urlStr string, requestHeader h
...
+		return nil, errors.New("proxy url function is nil")
+	}
+	// Generate the proxy URL and validate.
+	proxyURL, err := d.Proxy(req)
+	if err != nil {
+		return nil, err
+	}
+	if proxyURL == nil {
+		return nil, errors.New("generated proxy url is nil")
+	}
```
aojea, line 431: IIUIC this change the behavior, before if the returned `proxyURL` was nil meant "this connections should not be proxied" and now it will error , I think this should be just a noop
```suggestion
        // this request should not be proxied and use the original dialer
	if proxyURL == nil {
		return netDial, nil
	}
```

client.go
```diff
@@ -415,6 +414,53 @@ func (d *Dialer) DialContext(ctx context.Context, urlStr string, requestHeader h
...
+	if proxyURL.Scheme == "http" && d.NetDialTLSContext != nil {
+		return nil, errors.New("dial TLS function set for dialing non-HTTPS proxy")
+	}
```
aojea, line 434: This does not match the logic in 
```
	var netDial netDialerFunc
	switch {
	case u.Scheme == "https" && d.NetDialTLSContext != nil:
		netDial = d.NetDialTLSContext
	case d.NetDialContext != nil:
		netDial = d.NetDialContext
	case d.NetDial != nil:
		netDial = func(ctx context.Context, net, addr string) (net.Conn, error) {
			return d.NetDial(net, addr)
		}
	default:
		netDial = (&net.Dialer{}).DialContext
	}
```

that seems to indicate that `NetDialTLSContext` is ONLY used if the scheme is https , but ignored otherwise

client.go
```diff
@@ -415,6 +414,53 @@ func (d *Dialer) DialContext(ctx context.Context, urlStr string, requestHeader h
...
+	// For HTTPS proxy, either use TLS dial function set in Dialer (which performs
+	// the TLS handshake within the function), or create a new dial function by wrapping
+	// the passed "netDial" for the connection and using TLSClientConfig for the handshake.
+	if proxyURL.Scheme == "https" {
+		if d.NetDialTLSContext != nil {
+			netDial = d.NetDialTLSContext
```
aojea, line 440: note for self:

- url is http and proxy-url is https, we replace the dialer ... users can specify a custom tls dialer just for the proxy.
- url is https and proxy-url is https, the same tls dialer is used for both

proxy.go
```diff
@@ -29,7 +29,7 @@ func (fn netDialerFunc) DialContext(ctx context.Context, network, addr string) (
 }
 
 func proxyFromURL(proxyURL *url.URL, forwardDial netDialerFunc) (netDialerFunc, error) {
-	if proxyURL.Scheme == "http" {
+	if proxyURL.Scheme == "http" || proxyURL.Scheme == "https" {
```
aojea, line 32: is it there other possibility? doesn't this mean, we always use the `httpProxyDialer` ?

client.go
seans3: Fixed.

proxy.go
seans3: Yes--I'll always use the `httpProxyDialer`. The difference is the `forwardDial`, which now will also perform the TLS handshake if necessary.

client.go
seans3: > * url is https and proxy-url is https, the same tls dialer is used for both

I do *not* think this is correct. There is only one "dial". It is either to the "proxy" or to the "upstream" server. If there is a proxy, the dial is to the "proxy" (and the TLS handshake must occur in the dial). The "proxy" dials then dials the "upstream".  The second TLS handshake (not dial) to the "upstream" occurs over the previously established connection, and it MUST use the `TLSClientConfig`.

seans3:
> IIUIC the same certificate is used for both the proxy and the backend, don't we need to handle the case when both are different? what happens if the proxy requires a different certificate than the backend?

The client is responsible for configuring the root CA certificates. The client TLS config contains a cert pool, which can contain multiple root CA certificates. So for TLS to the proxy and the upstream, the TLS config root CA cert pool must contain CA's which have signed both the proxy and the upstream.

134130:
This PR also closes #950

client.go
```diff
@@ -426,7 +426,7 @@ func (d *Dialer) proxyNetDial(req *http.Request, netDial netDialerFunc) (netDial
 	if err != nil {
 		return nil, err
 	}
-	// Request should not be proxied; use originial dial function.
+	// Request should *not* be proxied; use originial dial function.
```
aojea, line 460: nit typo on `originial`

proxy.go
aojea: my point is that proxyFromURL now is just `return (&httpProxyDialer{proxyURL: proxyURL, forwardDial: forwardDial}).DialContext, nil` the other code will never be executed so we should not keep it if that is the case, or can lead to confusion

aojea: oh, good point, I was not considering those

client_proxy_server_test.go
```diff
@@ -0,0 +1,822 @@
...
+	proxyServer.StartTLS()
+	defer proxyServer.Close()
+	proxyServerURL, err := url.Parse(proxyServer.URL)
+	if err != nil {
+		t.Fatalf("error parsing websocket server URL: %v", err)
+	}
+	// Dial the websocket server to create the websocket connection,
+	// setting the proxy URL, the TLS CA data, and the requested subprotocol.
+	certPool := x509.NewCertPool()
+	certPool.AppendCertsFromPEM(websocketServerCert)
```
seans3, line 336: Add subtests for missing CA cert.

client_proxy_server_test.go
```diff
@@ -0,0 +1,822 @@
...
+	proxyServerURL, err := url.Parse(proxyServer.URL)
+	if err != nil {
+		t.Fatalf("error parsing websocket server URL: %v", err)
+	}
+	// Dial the websocket server to create the websocket connection through
+	// the proxy. The "NetDialTLSContext" function to dials the proxy and
+	// performs the TLS handshake. NOTE: Subsequent TLS handshake to backend
+	// (over proxied connection) uses TLSClientConfig for handshake.
+	certPool := x509.NewCertPool()
+	certPool.AppendCertsFromPEM(websocketServerCert)
```
seans3, line 595: Remove WebSocket self-signed cert.

client_proxy_server_test.go
```diff
@@ -0,0 +1,822 @@
...
+	dialer := Dialer{
+		Proxy: http.ProxyURL(proxyServerURL),
+		// Dial and TLS handshake function to proxy.
+		NetDialTLSContext: func(ctx context.Context, network, addr string) (net.Conn, error) {
+			netDialCalled.Add(1)
+			return tls.Dial(network, addr, tlsConfig)
+		},
+		// Used for second TLS handshake to backend server over previously
+		// established proxied connection.
+		TLSClientConfig: tlsConfig,
```
seans3, line 608: This should only be for WebSocket backend server.

client_proxy_server_test.go
```diff
@@ -0,0 +1,822 @@
+// Copyright 2025 The Gorilla WebSocket Authors. All rights reserved.
```
seans3, line 1: Add negative tests, where TLS fails when certs missing.

client.go
seans3: Fixed.

AlexVulaj:
Thanks for the PR @seans3 , this looks good to me!

AlexVulaj (approved):

seans3:
@AlexVulaj Thanks for the prompt review. Would it be appropriate to cut the next `v1.5.4` release? Again, thanks :)

liggitt:
probably want to merge the follow-up https://github.com/gorilla/websocket/pull/982 to deflake the unit test before cutting a tag

liggitt:
@AlexVulaj friendly ping on https://github.com/gorilla/websocket/pull/982 as a follow-up to this

reneleonhardt:
Hello @AlexVulaj other projects like k8s are relying on this project and had to switch from 1.5.3 to v1.5.4-0.20250319132907-e064f32e3674.

Would it be possible to tag a new release after a year of development?

briandavis-viz:
Hey @AlexVulaj, it would be great to have this in a released version, were you able to create a release tag so it can be referenced that way?

liggitt:
@AlexVulaj, sorry for nagging, it would be great to get the unit test fixup tweaks in https://github.com/gorilla/websocket/pull/982/files?w=1 merged and a release tagged if at all possible

Issue #826, opened 2022-11-03
[feature] concurrent safe write function 


0x-3FFF:
currently, the implementation of the write series function is concurrency-unsafe, it is recommended to introduce a concurrency-safe write function

canelohill:
Concurrent writes are the top support issue for this package.  

It's not useful to make the current methods threadsafe given the overall design of the API, but it might be worth doing to reduce the support burden.

jaitaiwan:
I've been thinking about this for awhile. Do we support thread-safeness into the future or do we just say "too bad" and ask folks to find a better library. I'm not entirely sure.

canelohill:
Drive the concurrency-related issues elsewhere by making the individual methods threadsafe. Support thread safety into the future.  

Developers of robust applications will want to handle concurrency within their application code. The current API is no hindrance to those developers and may be preferable to other APIs. 

I read through the examples in this repository and don't see how additional thread safety in the package will benefit them. Perhaps I need to spend more time looking at them. As an interesting side note, none of the examples use sync.Mutex.

hulkingshtick:
Given that deadline, compression and compression level are set by separate methods, it does not make sense to support concurrent calls to NextWriter, WriteMessage and WriteJSON. 

A previous comment suggests that the support burden can be reduced by making the individual methods thread-safe.  That's a lot of work for something that does not make sense. 

That said, there is some low-hanging fruit.  Based on a scan through the issues tracker, it looks like most of the concurrent writes are through WriteMessage and WriteJSON.  Adding a mutex to protect those two calls will probably eliminate most of the reported `concurrent write to websocket connection` panics and mostly do the right thing for the application.

hulkingshtick:
>   do we just say "too bad" and ask folks to find a better library.

No, tell them to fix their code by writing from a single goroutine or by using a mutex. 

You can also create a new higher-level package that encapsulates a common use pattern for the package, namely the read and write pump from the chat example (but not the chat example's hub).  This package can address concurrent write and other common needs such as liveness detection.  The package would not add protocol above the websocket layer or be applicable to all use cases.

Issue #957, opened 2024-09-02
[FEATURE] Reduce support issues related to concurrent writes


hulkingshtick:
### Is there an existing feature request for this?

- [X] I have searched the existing feature requests

### Is your feature request related to a problem? Please describe.

The [documentation says](https://pkg.go.dev/github.com/gorilla/websocket#hdr-Concurrency):

> Applications are responsible for ensuring that no more than one goroutine calls the write methods (NextWriter, SetWriteDeadline, WriteMessage, WriteJSON, EnableWriteCompression, SetCompressionLevel) concurrently

To help developers detect concurrency errors in their application, the connection makes a best effort to detect concurrent calls to WriteMessage, WriteJSON and Write on the writer returned from NextWriter. When a  concurrent calls is detected, the connection panics with the string `concurrent write to websocket connection`.  This panic is better than the alternatives such as commingling messages on the underlying network connection or a nil pointer exception.

The `concurrent write to websocket connection` panic is the most common support issue for the package.  The issue arises because developers incorrectly assume that concurrency is allowed or wrongly think they covered all the bases with regards to ensuring a single writer.

### Describe the solution that you would like.

Here are options for improving the situation:

### Improve the panic string

The terse panic string does not sufficiently describe the problem or a fix.  Create a markdown file on this repository explaining the concurrency restriction and suggested fixes.  Include a link to the markdown file in the panic string.

### Improve the detector

The [current detector](https://github.com/gorilla/websocket/blob/5e002381133d322c5f1305d171f3bdd07decf229/conn.go#L625-L635) is:

	if c.isWriting {
		panic("concurrent write to websocket connection")
	}
	c.isWriting = true

	err := c.write(w.frameType, c.writeDeadline, c.writeBuf[framePos:w.pos], extra)

	if !c.isWriting {
		panic("concurrent write to websocket connection")
	}
	c.isWriting = false

The second concurrent writer panics, but not the first.  The stack trace for the first concurrent writer may be the key to finding the bug in the application.  Cause the first writer to panic using the following code.  In this code, the field `c.writing` is an `int` and replaces the field `c.isWriting`.

	if c.Writing != 0 {
                c.Writing = 2
		panic("concurrent write to websocket connection")
	}
	c.Writing = 1

	err := c.write(w.frameType, c.writeDeadline, c.writeBuf[framePos:w.pos], extra)

	if c.Writing != 1 {
		panic("concurrent write to websocket connection")
	}
	c.Writing = 0

[...]

PR #980, opened 2025-03-09
Concurrent safe writes

Ale-Cas:

- [x] Feature

## Description

Implements concurrency-safe write functions, using a mutex, as suggested in https://github.com/gorilla/websocket/issues/826 and https://github.com/gorilla/websocket/issues/828

- Related Issues: 
  - https://github.com/gorilla/websocket/issues/826
  - https://github.com/gorilla/websocket/issues/828

## Added/updated tests?

- [x] Yes
      have not been included_

## Run verifications and test

- [x] `make test` is passing

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -250,10 +250,9 @@ type Conn struct {
 	writeBufSize  int
 	writeDeadline time.Time
 	writer        io.WriteCloser // the current writer returned to the application
-	isWriting     bool           // for best-effort concurrent write detection
-
-	writeErrMu sync.Mutex
-	writeErr   error
+	writeMux      *sync.Mutex
+	writeErrMu    sync.Mutex
+	writeErr      error
 
 	enableWriteCompression bool
 	compressionLevel       int
@@ -313,6 +312,7 @@ func newConn(conn net.Conn, isServer bool, readBufferSize, writeBufferSize int, 
 		writeBuf:               writeBuf,
 		writePool:              writeBufferPool,
 		writeBufSize:           writeBufferSize,
+		writeMux:               &sync.Mutex{},
 		enableWriteCompression: true,
 		compressionLevel:       defaultCompressionLevel,
 	}
@@ -377,8 +377,8 @@ func (c *Conn) read(n int) ([]byte, error) {
 }
 
 func (c *Conn) write(frameType int, deadline time.Time, buf0, buf1 []byte) error {
-	<-c.mu
-	defer func() { c.mu <- struct{}{} }()
+	c.writeMux.Lock()
+	defer c.writeMux.Unlock()
 
 	c.writeErrMu.Lock()
 	err := c.writeErr
@@ -622,17 +622,7 @@ func (w *messageWriter) flushFrame(final bool, extra []byte) error {
 	// concurrent writes. See the concurrency section in the package
 	// documentation for more info.
 
-	if c.isWriting {
-		panic("concurrent write to websocket connection")
-	}
-	c.isWriting = true
-
 	err := c.write(w.frameType, c.writeDeadline, c.writeBuf[framePos:w.pos], extra)
-
-	if !c.isWriting {
-		panic("concurrent write to websocket connection")
-	}
-	c.isWriting = false
 
 	if err != nil {
 		return w.endMessage(err)
@@ -750,15 +740,7 @@ func (c *Conn) WritePreparedMessage(pm *PreparedMessage) error {
 	if err != nil {
 		return err
 	}
-	if c.isWriting {
-		panic("concurrent write to websocket connection")
-	}
-	c.isWriting = true
 	err = c.write(frameType, c.writeDeadline, frameData, nil)
-	if !c.isWriting {
-		panic("concurrent write to websocket connection")
-	}
-	c.isWriting = false
 	return err
 }
 
--- a/conn_test.go
+++ b/conn_test.go
@@ -691,37 +691,21 @@ func TestUnexpectedCloseErrors(t *testing.T) {
 	}
 }
 
-type blockingWriter struct {
-	c1, c2 chan struct{}
-}
-
-func (w blockingWriter) Write(p []byte) (int, error) {
-	// Allow main to continue
-	close(w.c1)
-	// Wait for panic in main
-	<-w.c2
-	return len(p), nil
-}
-
-func TestConcurrentWritePanic(t *testing.T) {
-	w := blockingWriter{make(chan struct{}), make(chan struct{})}
-	c := newTestConn(nil, w, false)
-	go func() {
-		_ = c.WriteMessage(TextMessage, []byte{})
-	}()
-
-	// wait for goroutine to block in write.
-	<-w.c1
-
-	defer func() {
-		close(w.c2)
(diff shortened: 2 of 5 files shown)


hulkingshtick:
The PR does not enable concurrent writes.  

This PR replaces the best-effort race detector with a mutex.  The scope of the state that must be protected by a mutex is greater than the scope of the race detector.

The function TestConcurrentWrites does **not** call WriteMessage concurrently.   The tests will fail under the race detector when the test is modified to allow concurrent calls.

There is discussion in issues on why the Conn methods are not  not amenable to allowing concurrent writes.  This PR does not address any of those problems.

There are other problems with the PR, but I will not discuss them here because the points above are fatal.

Ale-Cas:
Hi @hulkingshtick, you're right this PR is not aiming to enable concurrent writes on the connection, the purpose was just to enable thread-safety in write functions using a mutex, and to remove the `concurrent write to websocket connection` panic.
It still looks like an improvement to me over the current `isWriting` boolean flag and it removes the burden to write some boiler plate code on the application, like the one we see in https://github.com/gorilla/websocket/issues/828#issue-1455062416.

Maybe I misunderstood your comment here https://github.com/gorilla/websocket/issues/826#issuecomment-2323461435:
> Adding a mutex to protect those two calls will probably eliminate most of the reported concurrent write to websocket connection panics and mostly do the right thing for the application.
Isn't this what you had in mind? 

> There are other problems with the PR, but I will not discuss them here because the points above are fatal.
If there are other problems, I'm open to feedback and happy to improve the PR. Otherwise if you still think it does not bring any value I'll close it up.

hulkingshtick:
> the purpose was just to enable thread-safety in write functions using a mutex

The concurrency issue is higher in the call stack.  Any fix for the problem will not require a mutex in `Conn.write`.

> and to remove the concurrent write to websocket connection panic

This panic is useful because the panic message explicitly tells the programmer what the problem is. The alternative is to panic in less obvious ways or to corrupt data.

> and it removes the burden to write some boiler plate code on the application side, like the one we see in https://github.com/gorilla/websocket/issues/828#issue-1455062416.

This PR does not relax the [concurrency requirements](https://pkg.go.dev/github.com/gorilla/websocket#hdr-Concurrency) because the PR does not fix the concurrent write problem. 
 
Update the your test to call WriteMessage concurrently and `go test -race` to see what happens.

> Isn't this what you had in mind?

I said that mutual exclusion lock in  `Conn.WriteMessage` and `Conn.WriteMessage` will reduce the number of issues reported against the package.  You added a lock lower in the call stack.

> If there are other problems, I'm open to feedback and happy to improve the PR.

- Don't include unrelated changes in a PR.
- Update comments when the referenced code changes.

hulkingshtick:
This test fails when run with `go test -race`.   This test starts all goroutines before waiting.  The test in the PR waits on the previously started goroutine before starting the next goroutine.

```
func TestConcurrentWrites(t *testing.T) {
	var connBuf bytes.Buffer
	c := newTestConn(nil, &connBuf, false)
	nGoroutines := 5
	done := make(chan error, nGoroutines)
	for i := 0; i < nGoroutines; i++ {
		go func() {
			err := c.WriteMessage(TextMessage, []byte{})
			done <- err
		}()
    }
    for i := 0; i < nGoroutines; i++ {
		err := <-done
		if err != nil {
			t.Fatal(err)
		}
	}
}
```

Ale-Cas:
I understand now, thank you very much for clarifying, I'm working on it and will follow your suggestions

Ale-Cas:
> Don't include unrelated changes in a PR.

Btw if you're referring to the changes in the files other than conn/conn_test, they've been included here just because I run `gofmt -w .` so those are just formatting changes, but if you want I can revert them

frankjkelly:
Very interested in this . . . . . following. . . .

Ale-Cas:
@hulkingshtick Thank you very much for the review and for providing that failing test!
I looked again at the issues and I noticed only now [your comment here](https://github.com/gorilla/websocket/issues/957#issuecomment-2329665095).

Looking again at both your comment and the code, I agree that there's no good way to make NextWriter thread-safe.

Still, I'm wondering if making the WriteMessage and WriteJSON methods thread-safe would reduce issues related to concurrent writes.
Additionally, it would make sense to improve the current detector with a counter, as you suggested, in the `conn.write` method to ensure it panics in the first concurrent call to NextWriter or `writer.Write`.

An alternative, [as you suggested](https://github.com/gorilla/websocket/issues/826#issuecomment-2323481052) would be to add a new package that encapsulates this common use case, which would provide developers with a robust higher-level API. 

Based on the outcome of this discussion, I will either update or close this PR.

hulkingshtick:
> Still, I'm wondering if making the WriteMessage and WriteJSON methods thread-safe would reduce issues related to concurrent writes.

Those methods are often on the stack in issues reporting the concurrent write panic, so yes, adding a mutex will cut down on reported issues. 

The goal of my suggestion is to reduce the support burden for the maintainers, not to make those applications correct.  I would hold off on any changes until the maintainers show that they are still interested in maintaining this package.

liggitt:
If the underlying Conn won't support concurrent writes well, is it a good idea to allow concurrency in calls to higher level methods? That seems like it will invite concurrent use that will be non-deterministic and cause more problems.

frankjkelly:
> If the underlying Conn won't support concurrent writes well, is it a good idea to allow concurrency in calls to higher level methods? That seems like it will invite concurrent use that will be non-deterministic and cause more problems.

It would be great to have concurrency but even if we can't if there were examples that showed recommended "templates" of  how to achieve concurrency (with locks? channels?) that would be great.

We have a use-case where we have multiple writers and we can figure out the locks but I'm sure whatever we will figure out will likely be suboptimal given our (current) shallow understanding of concurrency / goroutines / locking versus what the great folks here on in the open source world have. Either way, appreciative of the effort here by all sides.

ghost:
> If the underlying Conn won't support concurrent writes well, is it a good idea to allow concurrency in calls to higher level methods

The current code ensures that there are no concurrent writes to the underlying net.Conn.  

> if there were examples that showed recommended "templates" of how to achieve concurrency (with locks? channels?) that would be great.

https://github.com/gorilla/websocket/tree/main/examples

frankjkelly:
> > if there were examples that showed recommended "templates" of how to achieve concurrency (with locks? channels?) that would be great.
> 
> https://github.com/gorilla/websocket/tree/main/examples

Thanks @hulkingshtick but maybe I am missing something?

```
$ find . -name "*.go" | xargs grep -i Lock
$ 
```

I see some `<-` instances but not sure which one is specific what I am looking for. Can you advise?

hulkingshtick:
@frankjkelly  See [Conn.mu](https://github.com/gorilla/websocket/blob/main/conn.go#L247).

    // Write fields
    mu            chan struct{} // used as mutex to protect write to conn
    writeBuf      []byte        // frame is constructed in this buffer.
    ⋮

ghost:
@frankjkelly  The examples use goroutines and channels to ensure a single concurrent writer.  This approach is the simplest way to make an application robust against slow or dead peers.

Issue #314, opened 2017-12-19
Strange non blocking behavior of WriteMessage


EvgenySerg:
I'm tying to understand socket timeouts, but here i have a strange behavior of socket write. In this example of code there is no any timeouts, and we get looks like non blocking socket  in rotation write operations (hello1 and hello2 messages).
 If i make comment "//<part>" block of code, then i have desirable timeouts.

I don'understand this behavior.
How can i set different timeouts for different websocket.WriteMessage or Read operations?

Thanks.

 

```
package main

import (
	"log"
	"net/http"
	"time"
	"github.com/gorilla/websocket"
)

func main() {
	go Write()
	Listen()
}

func Write() {
	c, _, err := websocket.DefaultDialer.Dial("ws://127.0.0.1:8555", nil)
	if err != nil {
		log.Println(err.Error())
		return
	}
	err = c.SetWriteDeadline(time.Now().Add(time.Second * 2))
	if err != nil {
		log.Println(err.Error())
		return
	}

	for {
		err = c.WriteMessage(websocket.TextMessage, []byte("Hello 1"))
		if err != nil {
			log.Println(err.Error())
		}
		//<part>
		t := time.Second * 2
		log.Println(t)
		err = c.SetWriteDeadline(time.Now().Add(t))
		if err != nil {
			log.Println("asdasdsad")
			return
		}
		//</part>
		err = c.WriteMessage(websocket.TextMessage, []byte("Hello 2"))
		if err != nil {
			log.Println("c2 timeout " + err.Error())
		}

		time.Sleep(time.Second * 1)
	}
}

func handler(w http.ResponseWriter, r *http.Request) {
	upgrader := websocket.Upgrader{}
	c, err := upgrader.Upgrade(w, r, nil)
	if err != nil {
		log.Println(err.Error())
		return
	}
	for {
		_, data, err := c.ReadMessage()
		if err != nil {
			log.Println(err.Error())
		}
		log.Println(string(data))
		time.Sleep(time.Second * 5)
	}
}

func Listen() {
	http.HandleFunc("/", handler)
	http.ListenAndServe("127.0.0.1:8555", nil)
}

```

garyburd:
> How can i set different timeouts for different websocket.WriteMessage or Read operations?

Given desired timeout `t` for a write operation, call `c.SetWriteDeadline(time.Now().Add(t))` before the write operation.

[...]

PR #999, opened 2025-08-27
Update doc.go to mention Blocking

kylepl:

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [x] Documentation Update
- [ ] Go Version Update
- [ ] Dependency Update

## Description

https://github.com/gorilla/websocket/issues/314 mentions this explicitly, I found it handy to have this explicitly called out somewhere.

## Added/updated tests?

- [ ] Yes
- [x] No, and this is why: Just documentation


ghost:
The Go convention is to call out the cases where the API does not block, not the reverse as done here. 

For example, the [net.Conn Write](https://pkg.go.dev/net#Conn.Write) documentation does not state that the call blocks (as it must). 

Also, the condition is that the call blocks until the underlying network connection has read the application's data.

kylepl:
Ok, if that's the assumption.

Issue #994, opened 2025-07-18
[FEATURE] Option to control client's buffer for non-101 response body


thorrez:
### Is there an existing feature request for this?

- [x] I have searched the existing feature requests

### Is your feature request related to a problem? Please describe.

I have a websocket client that sometimes receives error HTTP responses (non-101) from servers. I want the client to be able to read a lot of the response body.

Currently, it can only read 1024 bytes of the response body, due to
https://github.com/gorilla/websocket/blob/e064f32e3674d9d79a8fd417b5bc06fa5c6cad8f/client.go#L367

### Describe the solution that you would like.

I would like an option to set that buffer to be larger than 1024 bytes.

Yurhigz:
Hello @thorrez , 

I have proposed a PR to deal with this kind of issue. Increasing the max buffer size up to 4096 bytes. I did not add another parameter but depending on what will be the comments on my PR I could consider adding one. 

You can find my PR there : https://github.com/gorilla/websocket/pull/1005

PR #1005, opened 2025-10-18
Client's buffer for non-101 response body issue optimization proposition

Yurhigz:

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Documentation Update
- [ ] Go Version Update
- [ ] Dependency Update

## Description

This PR addresses the buffer size limitation discussed in issue #994 (made by thorrez) by increasing maxErrorResponseSize from 1024 to 4096 bytes when handling non-101 WebSocket upgrade responses.
When a WebSocket upgrade request fails (returns a status code other than 101), the client reads a portion of the error response body to help application debugging. I wanted to increase the buffer size up to 4096 bytes from 1024 bytes. This limit increases allows larger error responses while still protecting against malicious intents (excessively large bodies).
I added 3 cases under the TestRespOnBadHandshake to verify correct behavior : 
- small error ( < 4096 bytes)
- large error ( > 4096 bytes)
- edge case ( = 4096 bytes)

- Related Issue #994 

## Added/updated tests?

- [x] Yes
      have not been included_

## Run verifications and test

- [x] `make test` is passing

### DIFF
--- a/client.go
+++ b/client.go
@@ -172,6 +172,7 @@ var nilDialer = *DefaultDialer
 // non-nil *http.Response so that callers can handle redirects, authentication,
 // etcetera. The response body may not contain the entire response and does not
 // need to be closed by the application.
+var maxErrorResponseSize = 4096
 func (d *Dialer) DialContext(ctx context.Context, urlStr string, requestHeader http.Header) (*Conn, *http.Response, error) {
 	if d == nil {
 		d = &nilDialer
@@ -364,9 +365,14 @@ func (d *Dialer) DialContext(ctx context.Context, urlStr string, requestHeader h
 		// Before closing the network connection on return from this
 		// function, slurp up some of the response to aid application
 		// debugging.
-		buf := make([]byte, 1024)
-		n, _ := io.ReadFull(resp.Body, buf)
-		resp.Body = io.NopCloser(bytes.NewReader(buf[:n]))
+
+		// Mon implémentation avec une maxErrorResponseSize
+		limReader := io.LimitReader(resp.Body, int64(maxErrorResponseSize))
+		buf, err := io.ReadAll(limReader)
+		if err != nil && err != io.EOF {
+			buf = []byte{}
+		}
+		resp.Body = io.NopCloser(bytes.NewReader(buf))
 		return nil, resp, ErrBadHandshake
 	}
 
--- a/client.go
+++ b/client.go
@@ -125,6 +125,11 @@ type Dialer struct {
 	// If Jar is nil, cookies are not sent in requests and ignored
 	// in responses.
 	Jar http.CookieJar
+
+	// MaxErrorBodySize specifies the maximum size of the error response buffer
+	// in the case of a bad hanshake. If zero, a defaut max size buffer of 1024 bytes
+	// is used.
+	MaxErrorBodySize int
 }
 
 // Dial creates a new client connection by calling DialContext with a background context.
@@ -172,8 +177,6 @@ var nilDialer = *DefaultDialer
 // non-nil *http.Response so that callers can handle redirects, authentication,
 // etcetera. The response body may not contain the entire response and does not
 // need to be closed by the application.
-var maxErrorResponseSize = 4096
-
 func (d *Dialer) DialContext(ctx context.Context, urlStr string, requestHeader http.Header) (*Conn, *http.Response, error) {
 	if d == nil {
 		d = &nilDialer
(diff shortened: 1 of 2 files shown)


client.go
```diff
@@ -172,6 +172,8 @@ var nilDialer = *DefaultDialer
 // non-nil *http.Response so that callers can handle redirects, authentication,
 // etcetera. The response body may not contain the entire response and does not
 // need to be closed by the application.
+var maxErrorResponseSize = 4096
```
ghost: The location of this variable declaration breaks the documentation (the DialContext documentation is no longer associated with the method).  Move the var declaration.

While you are at it, change it to a `const`.

client.go
```diff
@@ -364,9 +366,13 @@ func (d *Dialer) DialContext(ctx context.Context, urlStr string, requestHeader h
 		// Before closing the network connection on return from this
 		// function, slurp up some of the response to aid application
 		// debugging.
-		buf := make([]byte, 1024)
-		n, _ := io.ReadFull(resp.Body, buf)
-		resp.Body = io.NopCloser(bytes.NewReader(buf[:n]))
+
+		limReader := io.LimitReader(resp.Body, int64(maxErrorResponseSize))
```
ghost: Delete the conversion to `int64`. The conversion is not required if `maxErrorResponseSize` is changed from a variable to a constant.

client.go
```diff
@@ -364,9 +366,13 @@ func (d *Dialer) DialContext(ctx context.Context, urlStr string, requestHeader h
...
+		buf, err := io.ReadAll(limReader)
+		if err != nil && err != io.EOF {
```
ghost: io.ReadAll never returns io.EOF.  Use `if err != nil {`.

client_server_test.go
```diff
@@ -604,6 +606,76 @@ func TestRespOnBadHandshake(t *testing.T) {
...
+		if err != nil {
+			t.Fatalf("ReadAll(resp.Body) returned error %v", err)
+		}
+
+		resp.Body.Close()
+
+		if len(p) != maxErrorResponseSize {
+			t.Fatalf("body size=%d, want %d", len(p), maxErrorResponseSize)
+		}
+	})
```
ghost: Eliminate the duplicated test by using table driven test.

```	
for _, t := range []struct{ size, expect int }{{maxErrorResponseSize + 100, maxErrorResponseSize}, {maxErrorResponseSize, maxErrorResponseSize}} {
        body := bytes.Repeat([]byte{'a'}, t.size)
        ...
		if len(p) != t.expected {
			t.Fatalf("body size=%d, want %d for original body size %d", len(p), t.expected, t.size)
		}
 }
```

ghost (commented):
A simpler solution is to leave most of the code as is, but to get the size of the error response buffer from new Dialer field `MaxErrorBodySize int`.  If the `Dialer. MaxErrorBodySize` is zero, then use the original buffer size of 1024. 

The Dialer setting allows apps to adjust the value as needed.  The 4096 in this PR may not be large enough. 

Feel free to ignore everything I wrote as I am not a maintainer.

client.go
Yurhigz: Yup indeed you're right there !

Yurhigz:
> A simpler solution is to leave most of the code as is, but to get the size of the error response buffer from new Dialer field `MaxErrorBodySize int`. If the `Dialer. MaxErrorBodySize` is zero, then use the original buffer size of 1024.
> 
> The Dialer setting allows apps to adjust the value as needed. The 4096 in this PR may not be large enough.
> 
> Feel free to ignore everything I wrote as I am not a maintainer.

Well I thought about adding the MaxErrorBodySize as a setting within the Dialer Struct. But we must make sure it can’t read an arbitrarily large / unbounded error body, otherwise a malicious or just poorly configured server could return a huge response body and cause excessive memory usage or even a denial-of-service. That’s why I was thinking a hard cap (or a read timeout) would still be necessary.

ghost:
> But we must make sure it can’t read an arbitrarily large / unbounded error body

With my proposal, the maximum number of bytes read is neither arbitrary nor unbounded — it’s explicitly limited. The maximum is defined by the application, or defaults to 1024 bytes if not specified.

> otherwise a malicious

What specific threat are you referring to here? Are you suggesting a scenario where an attacker has modified the application’s source code?

> just poorly configured

It’s not the responsibility of the package to prevent an application from misconfiguring itself. Its role is to behave predictably within the constraints defined by the application.

> That’s why I was thinking a hard cap (or a read timeout) 

The current code has a [read timeout](https://pkg.go.dev/github.com/gorilla/websocket#Dialer.HandshakeTimeout).

Yurhigz:
Thank you for the feedback, I'll provide a better version by adding an element to the dial struct as you mentioned. With a default value at 0 which define a buffer size of 1024.

PR #1007, opened 2025-11-10
fix: avoid incrementing readErrCount for temporary errors and reset o…

yangdm0209:
### Commit title
fix: avoid incrementing readErrCount for temporary errors and reset on successful read

### Description

- Do not increment `readErrCount` for temporary errors such as timeouts, as they do not indicate actual connection failures.
- Reset `readErrCount` after successful read operations to reflect the true health status of the connection.
- This prevents `readErrCount` from growing due to normal timeout logic and provides a more accurate error state.

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Go Version Update
- [ ] Dependency Update

## Description

This PR addresses an issue where `readErrCount` was incremented on all errors, including temporary errors such as timeouts. Since timeouts often occur as part of normal connection management, especially when read deadlines are used, counting them as errors could lead to misleading high error counts and unnecessary connection termination or misreporting.

Now, only permanent errors cause `readErrCount` to increase, and a successful read resets this counter, better reflecting the real stability of the connection.

- Related Issue #
- Closes #  

## Added/updated tests?

- [ ] Yes
- [x] No, and this is why: _Please add details if you do not include tests (if the repo policy requires, plan test additions)_

## Run verifications and test

- [x] `make test` is passing

---

**Checklist (before submitting):**
- [x] Read the [Contributing guide](https://github.com/gorilla/.github/blob/main/CONTRIBUTING.md)
- [x] Read the [Code of Conduct](https://github.com/gorilla/.github/blob/main/CODE_OF_CONDUCT.md)
- [x] Provided clear commit messages and code comments.
- [x] Added/updated documentation as needed.
- [x] Requested review from gorilla/pull-request-reviewers.

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -1009,7 +1009,14 @@ func (c *Conn) NextReader() (messageType int, r io.Reader, err error) {
 		frameType, err := c.advanceFrame()
 		if err != nil {
 			c.readErr = hideTempErr(err)
+			if ne, ok := err.(net.Error); ok && (ne.Temporary() || ne.Timeout()) {
+				// temporary error, report it to the caller
+				return noFrame, nil, c.readErr
+			}
 			break
+		} else {
+			// reset on successful read
+			c.readErrCount = 0
 		}
 
 		if frameType == TextMessage || frameType == BinaryMessage {


ghost:
Although the underlying network connection may still be usable after returning an error, a `websocket.Conn` becomes unusable once its underlying connection encounters an error.

The `readErr` field reflects the health status of the `websocket.Conn`. It is set to a non-nil value when reads are no longer possible.

The `readErrCount` field tracks how many times the application has attempted to read from an unusable connection. Resetting this counter does not affect the actual health status of the connection.

None of the above prevents the normal use of a read deadline to detect when the peer is not sending data as expected.

Edit:  Because the websocket.Conn does call read on the underlying connection after a read error is encountered, the statement `c.readErrCount = 0` is only executed when `c.readErrCount` is zero.   The actual effect of this PR is that a temporary error may propagate to the application. The PR does not do what OP thinks it does.

PR #1013, opened 2026-02-17
Make compression reader/writer Close() idempotent

### LINKED ISSUE #859: [BUG] websocket on compress close duplicate

### Current Behavior

in this line 
https://github.com/gorilla/websocket/blob/ac0789be11725ab2285233e9a3800c2312cff4fc/compression.go#L138

and this line
https://github.com/gorilla/websocket/blob/ac0789be11725ab2285233e9a3800c2312cff4fc/conn.go#L1024

### Expected Behavior

if close duplicate will report this log 

` 2023/10/30 21:56:02 websocket: discarding reader close error: io: read/write on closed pipe`

### Steps To Reproduce

1. create websocket link with "EnableCompression"
2. use this websocket write someting
3. in server side got log "websocket: discarding reader close error: io: read/write on closed pipe"

### Anything else?

2023/10/30 21:56:02 websocket: discarding reader close error: io: read/write on closed pipe

veeceey:
Fixes #859

`flateReadWrapper.Read()` preemptively calls `Close()` when it encounters EOF to return the flate reader to the pool early. When `NextReader()` is called next, it calls `Close()` on the same wrapper again. This second `Close()` was returning `io.ErrClosedPipe`, causing spurious errors when compression was enabled.

The fix makes `Close()` on both `flateReadWrapper` and `flateWriteWrapper` idempotent — returning `nil` on an already-closed wrapper instead of an error. This follows Go conventions where closing an already-closed resource should be a safe no-op, and avoids the error path that was causing the log spam reported in #859.

The `Read()` and `Write()` methods still return their respective errors (`io.ErrClosedPipe` / `errWriteClosed`) when called on a closed wrapper, since those represent actual misuse.

Per [RFC 6455 Section 7.1.1](https://datatracker.ietf.org/doc/html/rfc6455#section-7.1.1), a close frame should only be sent once. The underlying issue was that the error from the redundant `Close()` call was being surfaced or logged, making it appear something was wrong when it was actually expected behavior.

Tests added for double-close scenarios on both read and write wrappers.

### DIFF
--- a/compression.go
+++ b/compression.go
@@ -108,7 +108,7 @@ func (w *flateWriteWrapper) Write(p []byte) (int, error) {
 
 func (w *flateWriteWrapper) Close() error {
 	if w.fw == nil {
-		return errWriteClosed
+		return nil
 	}
 	err1 := w.fw.Flush()
 	w.p.Put(w.fw)
@@ -143,7 +143,7 @@ func (r *flateReadWrapper) Read(p []byte) (int, error) {
 
 func (r *flateReadWrapper) Close() error {
 	if r.fr == nil {
-		return io.ErrClosedPipe
+		return nil
 	}
 	err := r.fr.Close()
 	flateReaderPool.Put(r.fr)
--- a/compression_test.go
+++ b/compression_test.go
@@ -77,3 +77,54 @@ func TestValidCompressionLevel(t *testing.T) {
 		}
 	}
 }
+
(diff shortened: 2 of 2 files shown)


veeceey:
hey, just checking in on this one - making Close() idempotent felt like the right thing to do. let me know if you'd prefer a different approach

PR #1014, opened 2026-02-27
Add conversion functions for writePoolData

koalabearguo:
Because writePoolData is not export,so it can not config WriteBufferPool field in Upgrader and Dialer

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Documentation Update
- [ ] Go Version Update
- [ ] Dependency Update

- Related Issue #
- Closes #

## Added/updated tests?

- [ ] Yes
      have not been included_

## Run verifications and test

- [x] `make test` is passing


mantidsboring:
Explain the situation where you need to know the type of the value in the pool.  Based on the documentation, the type is intentionally hidden.

koalabearguo:
the writePoolData struct is not exported,so when impl BufferPool interface Get/Put,it can not return writePoolData  type.

the bellow code line 517 use the writePoolData in conn.go ,it will translate the interface fail。
```
 512         mw.c = c
 513         mw.frameType = messageType
 514         mw.pos = maxFrameHeaderSize
 515 
 516         if c.writeBuf == nil {
 517                 wpd, ok := c.writePool.Get().(writePoolData)
 518                 if ok {
 519                         c.writeBuf = wpd.buf
 520                 } else {
 521                         c.writeBuf = make([]byte, c.writeBufSize)
 522                 }
 523         }
 524         return nil
```

koalabearguo:
Maybe you can give a demo for using BufferPool in Upgrader and Dialer,I can not run progam normal.

mantidsboring:
The pool implication stores values with type `any`.  The pool does not need to know the types of the values.   Per the documentation in this package, the [*sync.Pool](https://pkg.go.dev/sync#Pool) type can be used as a pool.  `*sync.Pool` is a good choice for most applicaitons.  Example use:

```
// create a pool
pool := &sync.Pool{}


// use it in the upgrader
u := websocket.Upgrader{
      ...
       WriteBufferPool: pool
}
```

A trivial implementation with a pool size of of one is:

```
package mypool

import (
	"sync/atomic"
)

type Pool struct {
	v atomic.Value
}

func (p *Pool) Put(v any) {
	p.v.Store(v)
}

func (p *Pool) Get() any {
	return p.v.Swap(nil)
}
```

Notice that the pool does not need to know the type of the value.

koalabearguo:
Thanks for getting back to me.

I use the bellow code to initial my program as you mentioned.
```
websocket.DefaultDialer.WriteBufferPool = &sync.Pool{}
```
but in conn.go bellow，the ok var is false forever，it forever log the "Error........."，the pool  is not used！
```
 517         if c.writeBuf == nil {
 518                 wpd, ok := c.writePool.Get().(writePoolData)
 519                 if ok {
 520                         c.writeBuf = wpd.buf
 521                 } else {
 522                         c.writeBuf = make([]byte, c.writeBufSize)
 523                         log.Println("Error........")
 524                 }
 525         }
 526         return nil
 527 }
```

koalabearguo:
ok，my program has only one byte pool，I dont want  websocket to new byteslice，maybe i can manage bytepool outside the websocket package

so this is my requirement,i need conversion functions for writePoolData,just new byteslice outside the package

mantidsboring:
Two things:

- I’m not a maintainer, but if I were, I’d want to see a use case that clearly cannot be solved without this PR.
- The maintainers haven’t merged a PR in almost a year, even though there are some high-quality submissions. I wouldn’t expect this to be merged anytime soon.

koalabearguo:
ok,i see.
thanks.
i has a bytepool in my main package,i just want to reuse the bytepool,this is the key point.
i just opt the memory usage for some docker env which is memory limited.

now i can use it local.so it dose not matter

thanks again.

PR #1029, opened 2026-09-14
client: copy TLS state onto DialContext response

### LINKED ISSUE #996: DialContext doesn't give back TLS information
Dear devs,

thank you for providing this library, however, I have an issue:

I have an TLS encrypted connection, but the response of DialContext() doesn't provide this informations as resp.TLS is always `nil`. :(

Here I dial: https://salsa.debian.org/mdosch/xmpp-dns/-/blob/gorilla/main.go#L956
There it is always nil: https://salsa.debian.org/mdosch/xmpp-dns/-/blob/gorilla/main.go#L1007

Best regards,
Martin

team-humaki:
Fixes #996

http.ReadResponse does not populate resp.TLS. Copy ConnectionState from the handshake conn so wss Dial/DialContext returns it, matching net/http.


jack-flippen:
#967 is another PR that address #996. 

Unlike this PR, #697 passes the TLS state to the trace callback.

